PR #11, opened 2016-02-09
Initial work on php support

kostajh:
I readily admit my regex skills are...limited. Any help you or others would like to add for building out PHP support would be most welcome! Thanks for this cool package.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -119,6 +119,14 @@
            :regex "class\\s*JJJ\\s*\\\(?"
            :tests ("class test(object):"))
 
+    ;; php
+    (:type "function" :language "php"
+           :regex "function\\s*JJJ\\s*\\\("
+           :tests ("function test()" "function test ()"))
+    (:type "variable" :language "php"
+           :regex "$\\s*JJJ\\s*=\\s*"
+           :tests ("$test = 1234"))
+
     ;; go
     (:type "variable" :language "go"
            :regex "\\s*\\bJJJ\\s*=\\s*" :tests ("test = 1234"))
@@ -165,6 +173,10 @@ and type to use for generating the grep command"
     (:language "javascript" :ext "js")
     (:language "javascript" :ext "jsx")
     (:language "javascript" :ext "html")
+    (:language "php" :ext "php")
+    (:language "php" :ext "module")
+    (:language "php" :ext "inc")
+    (:language "php" :ext "install")
     (:language "python" :ext "py")
     (:language "go" :ext "go"))
   "Mapping of programming lanaguage(s) to file extensions"


jacktasia:
Thanks for the PR! This looks pretty good, but I am going to try it with a few popular PHP projects before I merge it. Have you successfully used it? Looks like it's missing class property rules for example `$this->asdf = 1234;`

kostajh:
Thanks @jacktasia. Some more work needs to be done, for sure. Also, I had included `.module` and `.install` because 90% of Drupal code (a popular PHP framework) uses those extensions for PHP code.

I'm trying out the package on a file called `Comment.php` now and getting `Could not find rules for '.php file'`, any idea what's up with that?

jacktasia:
> I'm trying out the package on a file called Comment.php now and getting Could not find rules for '.php file', any idea what's up with that?

That makes it sounds like you haven't updated the package yet as that's the error it will give if an extension isn't mapped to language. Have you tried updating? If you're still having problems can you point me to the `Comment.php` file in question?

> Also, I had included .module and .install because 90% of Drupal code (a popular PHP framework) uses those extensions for PHP code.

So this means a drupal project will always have these types of files included in them? Is there a good drupal project I can test this with? Thanks for bringing this up!

kostajh:
@jacktasia I'm using use-package pointing to a local checkout of this project, definitely on the latest commit, and restarted Emacs to be sure. [Here's the file](https://github.com/savaslabs/squabble/blob/master/app/Comment.php) I was looking at.

For Drupal, have a look at https://ftp.drupal.org/files/projects/drupal-7.42.tar.gz and checkout `modules/system/system.module` for example.

jacktasia:
Thanks for info.

> I'm using use-package pointing to a local checkout of this project, definitely on the latest commit, and restarted Emacs to be sure. Here's the file I was looking at.

I don't get `Could not find rules for '.php file'` in that file when I try to jump, but also there doesn't seem like there's anything in that file to actually jump to (except for variable declarations in method signatures which would need a new rule, but should be super easy for me to add). I just get `'X'  php declaration not found.`...which is seemingly expected.

What exactly were you trying to jump to from that file? Also, to double check can you try` eval`ing `(prin1 dumb-jump-language-file-exts) and ensure `php` is in there? 

> For Drupal, have a look at https://ftp.drupal.org/files/projects/drupal-7.42.tar.gz and checkout modules/system/system.module for example.

Thanks. I'll check this out later.

kostajh:
@jacktasia I'm not sure when or how this happened, but the file extensions ended up in my `custom-set-variables`, and that excluded the rule for `.php` (but had the one for `.module` I had added before).

jacktasia:
@kostajh Ohhhhh, interesting. I guess that explains it. Thanks for reporting back! Is everything working as expected now?

kostajh:
Yes! Now I just need to work on adding a few more things :-)

Jack Angers writes:

> @kostajh Ohhhhh, interesting. I guess that explains it. Thanks for reporting back! Is everything working as expected now?
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/jacktasia/dumb-jump/pull/11#issuecomment-182673990


-- 
@kostajh | kosta@kostaharlan.net

jacktasia:
Awesome! If you point out specific use cases in a new issue then I would be more than happy to try to add the functionality.  For instance, "In repo X I want to be able to jump to symbol X in file Y when I am in file Z" or something like that :smile:

PR #14, opened 2016-02-13
add faust

magnetophon:
Nice package, thanks!

I'm no regex hero, but this seems to work for me.
The only issue I've noticed is that if there are multiple matches, it'll jump to a match without giving the menu to choose one of them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ There is currently basic support for the following languages:
 * Go
 * PHP
 * Ruby
+* Faust
 
 If you have any issues with the existing languages, or you want support for another one, then please open an issue. PRs are also welcome.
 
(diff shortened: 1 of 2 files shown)


jacktasia:
@magnetophon Thanks for the PR!

It looks like the test for your new rule is failing though. I know nothing about Faust, but it seems like it needs to be something like this:

`test(freq) = osc(freq) + 0.5;`

Also, would it make sense to add a  Faust `variable` rule too? As far as I can tell it seems like it could be exactly like Python's variable rule.

Thanks again!

magnetophon:
@jacktasia This should be better...

jacktasia:
@magnetophon Cool, thanks!

PR #20, opened 2016-04-26
C/C++ support

xianbaum:
Hey! I included some regex that adds support for C/C++ and includes definitions from the C preprocessor as well. I originally intended it to be for C support only, but I figured C++ support wouldn't be too hard to implement. (I was wrong about that. Regex is hard.) It seems to work very well in C. I tried it on the Lua source code and it worked extremely well. I also tried it on the Firefox source code, and it hangs for about a minute and fails. It might have something to do with the fact that the Firefox source code is huge.

There actually is a potential problem with this. When I run the tests, it says it fails 3 of 70 tests. I don't know what could be wrong with it, and the errors don't mean much to me... Could it have something to do with the length of my regex? I tried to write regex for just about every case in C/C++. I noticed that some of the regex above seemed to be split up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ There is currently basic support for the following languages:
 * Lua
 * Rust
 * CoffeeScript
+* Perl
 
 If you have any issues with the existing languages, or you want support for another one, then please [open an issue](https://github.com/jacktasia/dumb-jump/issues). PRs are also welcome. If you'd like to add a language these PRs for [lua](https://github.com/jacktasia/dumb-jump/pull/33) and [rust](https://github.com/jacktasia/dumb-jump/pull/57) are good examples.
 
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -262,6 +262,16 @@
            :regex "\\bJJJ\\s*<-\\s*function"
            :tests ("test <- function"))
 
+
+    ;; perl
+    (:type "function" :supports ("ag" "grep") :language "perl"
+           :regex "sub\\s*JJJ\\s*\\\{"
+           :tests ("sub test{" "sub test {"))
+
+    (:type "variable" :supports ("ag" "grep") :language "perl"
+           :regex "JJJ\\s*=\\s*"
+           :tests ("$test = 1234"))
+
     ;; php
     (:type "function" :supports ("ag" "grep") :language "php"
            :regex "function\\s*JJJ\\s*\\\("
@@ -478,6 +488,11 @@ and type to use for generating the grep command"
     (:language "javascript" :ext "js" :agtype "js")
     (:language "javascript" :ext "jsx" :agtype "js")
     (:language "javascript" :ext "html" :agtype "html")
+    (:language "perl" :ext "pl" :agtype "perl")
+    (:language "perl" :ext "pm" :agtype "perl")
+    (:language "perl" :ext "pm6" :agtype "perl")
+    (:language "perl" :ext "pod" :agtype "perl")
+    (:language "perl" :ext "t" :agtype "perl")
     (:language "php" :ext "php" :agtype "php")
     (:language "php" :ext "inc" :agtype "php")
     (:language "ruby" :ext "rb" :agtype "ruby")
@@ -498,6 +513,7 @@ and type to use for generating the grep command"
     (:language "javascript" :type "variable" :right "^\\." :left nil)
     (:language "javascript" :type "variable" :right "^;" :left nil)
     (:language "php" :type "function" :right "^(" :left nil)
+    (:language "perl" :type "function" :right "^(" :left nil)
     (:language "elisp" :type "function" :right nil :left "($")
     (:language "elisp" :type "variable" :right "^)" :left nil))
 


jacktasia:
Thanks for the PR! It will be great to have C/C++ support.

> There actually is a potential problem with this. When I run the tests, it says it fails 3 of 70 tests. I don't know what could be wrong with it, and the errors don't mean much to me...

Yeah, it looks like some of your regex's `:tests` and/or `:not` are failing. Right now, as you noticed, it doesn't surface exactly which regex or regex test case failed. I usually just add in a logging statement when I run into this, but the time has come for the test code to be changed up so a full report is returned with _exactly_ which regex are failing for each `:tests` or `:not`.

I will make this improvment tonight after work and report back on exactly what your issue is and have code that you can pull down so you can run the reports yourself too.

Thanks again for the PR!

jacktasia:
It looks like the regexes are failing when the tests run with `grep`. If you want you can find out more about which tests are failing by [editing the test file](https://github.com/jacktasia/dumb-jump/blob/v0.3.9/test/dumb-jump-test.el#L184) with something like:
    `(when rule-failures
       (prin1 rule-failures))`

I really want C/C++  support so this inspired me to think that rules should be [tied to a search program](https://github.com/jacktasia/dumb-jump/issues/23). This would allow dumb-jump not to be forced to have regexes that work in both `ag` and `grep`, which will also make supporting other search programs easier.

I understand if you don't want to do the work to support `grep` too,  but when I get around to #23, I will take an another look at this PR and see if we would then only need to mark these rules as `ag`-only.

Thanks again!

xianbaum:
You're welcome! Thank you so much for looking into that for me. I was stumped by the failed tests. I just assumed that it worked in grep, too. I'll try my best to adapt the regex so that it works in both grep and ag.

wasamasa:
Minor nitpick: It's "c++", not "cpp" (which stands for the C preprocessor).

@xianbaum Any progress on this one?  C is pretty much the only thing I'm still using etags for, I'd like trying out dumb-jump if it had support for that.

xianbaum:
No progress, but I'll work on it today.

jacktasia:
@xianbaum #23 is now in master.  So you should only have to add `:supports ("ag")` to all the new rules in this PR now.

xianbaum:
Oh, thank you! I'm working on a regex that works with both grep and ag. I've just been looking for cases where they don't work and trying to fix it. The biggest problem I have is, I don't think I can differentiate "return x" and "my_struct x;" without a lookarounds, which grep doesn't support unless perl-style regex are enabled. There are probably more cases, like forward declarations ("class MyClass;"). and that would also be picked up by the and that might also be interpreted as a variable in this regex. Is it possible to use a different regex for grep and ag? Other than that, I'm just trying to improve the regex and I'm looking for cases where it doesn't find anything.

jacktasia:
> Is it possible to use a different regex for grep and ag?

Yes, just have one rule that is `:supports ("ag")` and one that is `:supports ("grep")`. That said, I wouldn't worry too much about supporting `grep`. I think most people use `ag`.

xianbaum:
Hey! Sorry I'm so slow with this. I'm actually having a little trouble with this and I think I need help. For some reason, I can't get my regex to work with ag on the latest version, but the grep regex is working. In 5313ef651b58dd9b8b9fcb388856b8dcbf1b791b, the same regex is working for me with ag. (Side-note: when I run tests, it often gets stuck after test 2 for some reason, but not all the time. I'm not sure why, but it's probably something with my machine). It just says "declaration not found" when I have ag installed, but when I uninstall it, it works with grep. I think this passes all the tests, (I'm not certain since when I try to run it it hangs on test 2, but I think I remember it pasisng before). I'm not sure if this is the best way to share my code, but here is what I put in the latest commit (193e47cba2d69878b2081f4c01edcf7d235ea744)

In dumb-jump-find-rules:
```elisp
    (:type "function" :supports ("ag" "grep") :language "c++"
	   :regex "((\\w|>)(\\s+|(\\s*(\\*+|\\\(+)\\s*|\\s*::)+)?JJJ(\\s*|(\\s*\\\)+\\s*)+)\\\([ a-zA-Z_\\s,&*]*(\\.\\.\\.)?\\s*(\\\)|$)\\s*(const\\s*)?(->[ a-zA-Z_\\s)\\\(\\\)\\.]*)?(\\{|$)|typedef\\s+[ a-zA-Z_\\s\\\(\\*]+JJJ[\\\) \\s]*\\\()|#\\s*define\\s+JJJ\\\("
	   :tests ("int test(){" "char **test( char * arg ){" "int (*test)() {" "auto Person::test (Builder &reference ) -> delctype( builder.func() ) {" "test::test( int *random_argument) const {")
	   :not ("return test();" "int test( a, b );" "if( test() ) {"))
    (:type "variable" :supports ("grep") :language "c++"
	   :regex "\\b(\\w+|,)\\s[ \\s*&]*JJJ\\s*(\\[[0-9\\s ]*\\]\\s*)*\\s*(=|;|,|\\\))|#define\\s+JJJ\\b"
	   :tests ("int test=2;" "char *test;" "int x = 1, test = 2" "int test[20];" "#define test"))
    (:type "variable" :supports ("ag") :language "c++"
	   :regex "\\b(?!(class\\b|struct\\b|return\\b))(\\w+|,)\\s[ \\s*&]*JJJ\\s*(\\[[0-9\\s ]*\\]\\s*)*\\s*(=|;|,|\\\))|#define\\s+JJJ\\b"
	   :tests ("int test=2;" "char *test;" "int x = 1, test = 2" "int test[20];" "#define test" "typedef int test;")
	   :not ("return test;" "#define NOT test" "else test=2;"))
    (:type "type" :supports ("ag" "grep") :language "c++"
	   :regex "\\b(class|struct|enum|union)\\b\\s*JJJ\\b\\s*(final\\s*)?(:((\\s*\\w+\\s*::)*\\s*\\w*\\s*<?(\\s*\\w+\\s*::)*\\w+>?\\s*,*)+)?((\\{|$))|}\\s*JJJ\\b\\s*;"
	   :tests ("typedef struct test {" "enum test {" "} test;" "union test {" "class test final: public Parent1, private Parent2{" "class test : public std::vector<int> {")
	   :not("union test var;" "struct test function() {"))
```
in dumb-jump-language-file-exts:
```elisp
    (:language "c++" :ext "c" :agtype "c++")
    (:language "c++" :ext "h" :agtype "c++")
    (:language "c++" :ext "cpp" :agtype "c++")
    (:language "c++" :ext "hpp" :agtype "c++")
    (:language "c++" :ext "cxx" :agtype "c++")
    (:language "c++" :ext "hxx" :agtype "c++")
    (:language "c++" :ext "cc" :agtype "c++")
    (:language "c++" :ext "hh" :agtype "c++")
    (:language "c++" :ext "c++" :agtype "c++")
    (:language "c++" :ext "h++" :agtype "c++")
```

The same regex works on my machine ag with commit 5313ef651b58dd9b8b9fcb388856b8dcbf1b791b with this code (with test errors because it doesn't work with grep);

In dumb-jump-find-rules:
```elisp
    (:type "function" :language "c++"
	   :regex "((\\w|>)(\\s+|(\\s*(\\*+|\\\(+)\\s*|\\s*::)+)?JJJ(\\s*|(\\s*\\\)+\\s*)+)\\\([ a-zA-Z_\\s,&*]*(\\.\\.\\.)?\\s*(\\\)|$)\\s*(const\\s*)?(->[ a-zA-Z_\\s)\\\(\\\)\\.]*)?(\\{|$)|typedef\\s+[ a-zA-Z_\\s\\\(\\*]+JJJ[\\\) \\s]*\\\()|#\\s*define\\s+JJJ\\\("
	   :tests ("int test(){" "char **test( char * arg ){" "int (*test)() {" "auto Person::test (Builder &reference ) -> delctype( builder.func() ) {" "test::test( int *random_argument) const {")
	   :not ("return test();" "int test( a, b );" "if( test() ) {"))
    (:type "variable" :language "c++"
	   :regex "\\b(?!(class\\b|struct\\b|return\\b))(\\w+|,)\\s[ \\s*&]*JJJ\\s*(\\[[0-9\\s ]*\\]\\s*)*\\s*(=|;|,|\\\))|#define\\s+JJJ\\b"
	   :tests ("int test=2;" "char *test;" "int x = 1, test = 2" "int test[20];" "#define test" "typedef int test;")
	   :not ("return test;" "#define NOT test" "else test=2;"))
    (:type "type" :language "c++"
	   :regex "\\b(class|struct|enum|union)\\b\\s*JJJ\\b\\s*(final\\s*)?(:((\\s*\\w+\\s*::)*\\s*\\w*\\s*<?(\\s*\\w+\\s*::)*\\w+>?\\s*,*)+)?((\\{|$))|}\\s*JJJ\\b\\s*;"
	   :tests ("typedef struct test {" "enum test {" "} test;" "union test {" "class test final: public Parent1, private Parent2{" "class test : public std::vector<int> {")
	   :not("union test var;" "struct test function() {"))
```
in dumb-jump-language-file-exts:
```elisp
    (:language "c++" :ext "c")
    (:language "c++" :ext "h")
    (:language "c++" :ext "cpp")
    (:language "c++" :ext "hpp")
    (:language "c++" :ext "c++")
    (:language "c++" :ext "h++")
    (:language "c++" :ext "cxx")
    (:language "c++" :ext "hxx")
    (:language "c++" :ext "cc")
    (:language "c++" :ext "hh")
```
Again, sorry it took so long to get back only to ask for help... I was actually working on the older commit to test the regex and I am thinking maybe there is something I am missing when I adapted it to the later commits. For the later commits, all I had to do was add the `:agtype` keyword and the `:supports` keyword, right? Thank you!

jacktasia:
No worries! So your issue is probably your `agtype` for `dumb-jump-language-file-exts`. It should be `cpp` and not `c++` as [shown here](https://github.com/ggreer/the_silver_searcher/blob/9fc8924c0b3d9c8ab088a56c528f3a9bf1875836/tests/list_file_types.t#L42)

As far as the test running stalling. You probably want to kill the `.cask` directory and re-install:

```
cd dumb-jump
rm -rf .cask
cask
make test
```

Please let me know if you still run into issues.

jacktasia:
FWIW, my most recent commit should make it easier to determine exactly which regex/test failed when testing (`make test`).

xianbaum:
Thanks!! I'll definitely work on it this weekend

xianbaum:
Hey! I was having problems where my regex wasn't working but I realized agtype is case sensitive. I double checked and realized ag's agtype's cpp has lists the extension for .C but not .c:

```
--cpp
    .cpp .cc .C .cxx .m  .hpp  .hh .h .H .hxx
```

But I also noticed it has lower-case C under cc, so I changed the agtype to cc (is that alright?)

```
--cc
    .c .h .xs
```

I was stumped for a while on why a lot of my searches weren't successful much of the time. They ended up working when I changed the agtype to cc.

Anyway, I'll probably update this PR with my new code tomorrow night, once I double check everything!

jacktasia:
@xianbaum Awesome. Thanks for working on this! Please let me know if you need any help.

PR #21, opened 2016-04-26
introduce dumb-jump-after-jump-hook

jordonbiondo:
Introduces a hook run after jumping. 

Here is my personal use case but it could be used for a variety of purposes.

```lisp
  (add-hook 'dumb-jump-after-jump-hook
            (defun jordon-dumb-jump-pulse-line ()
              (pulse-momentary-highlight-one-line (point))))
```

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -259,6 +259,11 @@ immediately to the right of a symbol then it's probably a function call"
   "The default project to search within if a project root is not found"
   :group 'dumb-jump)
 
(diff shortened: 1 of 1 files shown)


jacktasia:
Thanks for the PR! This is a very good idea. I am currently using [beacon](https://github.com/Malabarba/beacon) for this kind of functionality, but I think this is better since it opens the door to other uses too.

Let me try this out tonight after work. I'll report back and/or merge then. Thanks again!

jordonbiondo:
Thanks!

PR #39, opened 2016-05-27
added support for limiting ag search to file types

anuragpeshne:
while search for a certain function in a Python file, dumb-jump got me to function with same name in JavaScript file. Hence I added support for limiting search to file types which should speed up searching too.


anuragpeshne:
Got some errors, I'll fix it and open a new request.

Issue #38, opened 2016-05-27
Extensions for clojure variants


ejlo:
Please add the extensions for commonly used variants of clojure: .cljs and .cljc.

jacktasia:
Thanks for reporting this. I will add this into the next release for sure.

jacktasia:
@ejlo Done now. This update should be available via MELPA within ~12 hours. Thanks again for reporting this.

PR #40, opened 2016-05-27
added support for limiting ag search to file types

anuragpeshne:
while search for a certain function in a Python file, dumb-jump got me to function with same name in JavaScript file. Hence I added support for limiting search to file types which should speed up searching too.

Corrected the way ag file types are concatenated.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -809,9 +809,9 @@ denoter file/dir is found or uses dumb-jump-default-profile"
 
 (defun dumb-jump-get-ag-type-by-language (language)
   "Returns list of ag type argument for a LANGUAGE"
-  (--map (plist-get it :agtype)
-         (--filter (string= (plist-get it :agtype) language)
-                  dumb-jump-language-file-exts)))
+  (-distinct (--map (plist-get it :agtype)
+                    (--filter (string= (plist-get it :agtype) language)
+                              dumb-jump-language-file-exts))))
 
 (defun dumb-jump-get-rules-by-language (language)
   "Returns a list of rules for the LANGUAGE"
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -776,9 +776,7 @@ denoter file/dir is found or uses dumb-jump-default-profile"
                       (if (s-ends-with? ".gz" cur-file)
                           " --search-zip"
                         "")
-                      (dumb-jump-arg-joiner
-                       "--"
-                       agtypes)))
+                      (s-join "" (--map (format " --%s" it) agtypes))))
          (exclude-args (dumb-jump-arg-joiner "--ignore-dir" (--map (s-replace proj "" it) exclude-paths)))
          (regex-args (format "\"%s\"" (s-join "|" filled-regexes))))
     (if (= (length regexes) 0)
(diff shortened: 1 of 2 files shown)


anuragpeshne:
The code is failing just one test case
`dumb-jump-generate-ag-command-no-ctx-test`
Probably because of the extra whitespace added in the `cmd` generated

jacktasia:
Thanks for the PR! This is an interesting idea. Have you noticed an increase jump speed?

It looks like the `dumb-jump-generate-ag-command-no-ctx-test` test is failing because it hasn't been updated to to include the `--elisp` argument. Additionally it seems like you'll want to wrap `agtypes` with `-distinct` so it doesn't show up more than once.

image

anuragpeshne:
adding `-distinct` did the trick! thanks.
I have also updated the test file to expect *ag-types*.

I tried benchmarking the scripts but couldn't notice anything remarkable, maybe it will be prominent in a large folder.

jacktasia:
Awesome, thanks for fixing that test. Looks like there's just there's just a tiny merge conflict now. Should be trivial to fix because I think it happened when I resolved #38, which was just adding two new clojure extensions. Thanks again!

scottaj:
I use dumb jump with JavaScript and AngularJS. One of the best things about it is that if I use dumb jump on a variable or function in an HTML template, it will find it in the associated JS file. Will this change break that functionality?

jacktasia:
@scottaj Fantastic point. Thank you so much for bringing this up. Based on [this](https://github.com/ggreer/the_silver_searcher/blob/9fc8924c0b3d9c8ab088a56c528f3a9bf1875836/tests/list_file_types.t#L111) and [this](https://github.com/ggreer/the_silver_searcher/blob/9fc8924c0b3d9c8ab088a56c528f3a9bf1875836/tests/list_file_types.t#L99). I think this PR would indeed break the functionality you're talking about. So @anuragpeshne I think the easiest solution is to wrap this in a flag like `dumb-jump-use-agtypes` that defaults to `nil`.  This way if someone has a huge non-mixed language project they can easily turn on this functionality for faster jumping, but it won't hurt people who depend on the current functionality. 
```
(defcustom dumb-jump-use-agtypes
  nil
  "When enabled dumb-jump will only search the same filetypes as the starting file."
  :group 'dumb-jump)
```

Thanks again!

anuragpeshne:
@scottaj I had this case in my mind while limiting the search and hence we pass list of file types to ag. In case of [Javascript](https://github.com/anuragpeshne/dumb-jump/blob/master/dumb-jump.el#L311-L313) we pass both `--js` and `--html`.
`RUNNING CMD '"ag --nocolor --nogroup --js --html \"(service|factory)\\((['...`
This would not work if there are more than one language in a single extension (example: PHP emitting Javascript). But that would anyways won't work, since we'll be using PHP's regex to grep JS variables/functions. If there are any other cases, please let me know.

@jacktasia There was a small bug while extracting `agtype` from `dumb-jump-language-file-exts` which is now resolved. Also, I've added `agtype` for scala which now resolves the conflict with base branch.

jacktasia:
@anuragpeshne Awesome, thanks! Let me play with this a bit before I merge it.

jacktasia:
@anuragpeshne Looks good to me. Thanks again!

@scottaj I tested this PR jumping from HTML to javascript and it worked just fine. Please open an issue if you notice anything. Thanks again!

Issue #23, opened 2016-04-27
Add `:supports` property to rules


jacktasia:
So we can set which search program (`ag` or  `grep`) the rule supports. Stopping the need for a rule to support `grep` too, which often makes the regex way more complicated. Also opens the door to support other searchers like `pt` in interesting ways.

PR #46, opened 2016-06-06
Add ruby's class and module support

wamaral:
Adds support for the following constructions in Ruby:

* `class Foo`

* `module Foo`

This depends on #23 since grep does not support these regexes


jacktasia:
#23 was merged in so you should only need to add  `:supports ("ag")` for all your new rules

jacktasia:
Thanks!

PR #47, opened 2016-06-06
Add extra matches for Clojure

wamaral:
Adds support for the following constructions in Clojure:

* `(def foo`

* `(defn foo` / `(defn- foo`

* `(defmacro foo`

* `(deftype foo`

* `(defmulti foo`

* `(defmethod foo`

* `(definterface foo`

* `(defprotocol foo`

* `(defrecord foo`

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -134,10 +134,42 @@
            :not ("(defun blah (test-1)" "(defun blah (test-2 blah)" "(defun (blah test-3)"))
 
     ;; clojure
-    (:type "function" :language "clojure" :regex "\\\(defn-?\\s+JJJ\\j"
+    (:type "variable" :language "clojure"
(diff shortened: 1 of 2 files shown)


dumb-jump.el
```diff
@@ -129,15 +129,46 @@
 
     ;; variable in method signature
     (:type "variable" :language "elisp"
-           :regex "\\(defun\\s*.+\\\(?\\s*JJJ\\j\\s*\\\)?"
+           :regex "\\\(defun\\s*.+\\\(?\\s*JJJ\\j\\s*\\\)?"
```
wamaral: :gift:

jacktasia:
#23 was merged in so you should only need to add  `:supports ("ag")` for all your new rules

jacktasia:
Thanks!

PR #48, opened 2016-06-06
Add zsh support

abend:
Hi.  Thanks for this great package!  

I had trouble running the tests using zsh.  It appears to be a difference in how bash and zsh handle single quotes inside single quoted strings.  I think bash is a little more lenient.

This fix allows the tests to pass under zsh, bash, and sh.  Csh and ksh still have problems, but since I don't use them, I haven't put any effort into them.

If you like, I can add a feature that runs the tests under different shells (if installed).  I wanted to keep this patch small though. 

Let me know if you'd like anything changed.

Thanks!


jacktasia:
Awesome, thanks for doing this!

Issue #55, opened 2016-07-01
Not usable in shell


zenspider:
I'll often be investigating things inside an emacs shell. Either running tests or grepping around or whatever. From there I'd like to be able to put point on a symbol and invoke dumb-jump. But because the buffer is in shell-mode, it fails out. It'd be nice to be able to set a "default language" or something to let it at least attempt a jump.

jacktasia:
Thanks for the report. So we also need to know the project directory to search. One idea would be if a buffer is `*eshell*`(?) we somehow deduce the project path (via `cwd`?) and either get the language from the `.dumbjump` file in that project dir or by prompting the user with a list of languages. Thoughts?

zenspider:
On July 1, 2016 at 11:36:39, Jack Angers (notifications@github.com) wrote:
> Thanks for the report. So we also need to know the project directory to search. One idea  
> would be if a buffer is `*eshell*`(?) we somehow deduce the project path (via `cwd`?)  
> and either get the language from the `.dumbjump` file in that project dir or by prompting  
> the user with a list of languages. Thoughts?


I use regular `M-x shell`, not eshell. But you can always use `default-directory` no matter what shell / buffer / mode you’re in.

jacktasia:
Great point. I've thought this over a bit, and I think I can add support for this pretty easily. I will try to get something working over the weekend.

I am thinking of trying to determine the programming language automatically by the contents of the shell buffer, because if you're jumping from like a stacktrace, it will almost certainly reference files with their extension.

zenspider:
I would be happy to put something in a directory local variable too.

jacktasia:
@zenspider This should be available via MELPA now. Please open a new issue if you have any issues or  have other ideas. Thanks again!

jacktasia:
@zenspider Have you had a chance to try this yet? I am curious if it works well for you or not. Thanks.

PR #56, opened 2016-07-02
add support for jumping shell buffers #55

jacktasia:
As discussed in #55

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -551,7 +551,6 @@ denoter file/dir is found or uses dumb-jump-default-profile"
     (dumb-jump-issue-result "nosymbol"))
    (t
     (dumb-jump-fetch-file-results))))
-;    (dumb-jump-fetch-results))))
 
 (defun dumb-jump-fetch-shell-results ()
   (let* ((cur-file (buffer-name))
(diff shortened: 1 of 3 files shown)


jacktasia:
@zenspider This seems to work for me. Can you give this a try a let me know if you have any feedback before I merge?

jacktasia:
When running from a shell buffer, it will first check for a `.dumbjump` file in the project root with a line containing just `language X` (like `language python`, `language Ruby` - there should only be one). If that doesn't exist then, it will attempt to guess based on the file extensions it finds in the buffer contents. Right now the latter is problematic if you do `ls` in a directory with lots of different extensions, as it could accidently use the wrong language.

zenspider:
I wasn't sure how to test this properly as I currently have it installed via package. I used `curl` to replace the file and then bytecompiled it. It isn't working for me tho, as I'm currently getting: `"Please save your file before jumping."` in response to using it in `M-x shell`.

jacktasia:
@zenspider Sorry, here's some instructions:

1.  Select all copy/paste the [new code here](https://raw.githubusercontent.com/jacktasia/dumb-jump/47077940e891b21b8a13ea63744f95e541866c6b/dumb-jump.el) into a fresh buffer in emacs 
2. <kbd>M-x</kbd> `eval-buffer` on the buffer with the new code.

Please let me know if you still have problems. Thanks.

PR #59, opened 2016-07-07
Add autoload cookies to three interactive commands

10sr:
I want to use these commands without explicitly requiring the library XD


jacktasia:
Thanks!

10sr:
Thank you!

Issue #28, opened 2016-05-26
Support for CoffeeScript would be nice


mck-:

jonboiser:
Is there a general strategy for adding a new language? I was reading through the code and it looks like you just add a handful of plists for different symbol types in `dumb-jump-find-rules`, then some other variables in the other `defcustoms`.

Do you think you could give a brief checklist of what needs to be done? I can give it a try with coffeescript and write it up as a guide for contributors.

jacktasia:
@jonboiser 

Yes, multiple languages have been added via PR. Some good examples are [lua](https://github.com/jacktasia/dumb-jump/pull/33) and [rust](https://github.com/jacktasia/dumb-jump/pull/57) I've been meaning to [add this to the README](https://github.com/jacktasia/dumb-jump/commit/513a432569a26e10a954ca274b2aa0baeaaed656). Thanks for reminding me. Please let me know if I can help with adding coffeescript.

PR #61, opened 2016-07-14
[WIP] Add coffeescript support

jonboiser:
#28 

This is a work in progress, starting with the regex for functions.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -170,6 +170,23 @@
            :regex "\\(defrecord\\s+JJJ\\j"
            :tests ("(defrecord test [foo]"))
 
+    ;; coffeescript
+    (:type "function" :supports ("ag") :language "coffeescript"
+           :regex "^(?!#)\\s*JJJ\\s*[=:].*[-=]>"
+           :tests ("test = ()  =>" "test= =>" "test = ->" "test=()->"
+                   "test : ()  =>" "test: =>" "test : ->" "test:()->")
+           :not ("# test = =>" "test = 1"))
+
+    (:type "variable" :supports ("ag") :language "coffeescript"
+           :regex "^(?!#)\\s*JJJ\\s*[:=][^:=-](?!.*>)"
+           :tests ("test = $" "test : [" "test = {" "test = a")
+           :not ("test::a" "test: =>" "test == 1" "# test = 1"))
+
+    (:type "class" :supports ("ag") :language "coffeescript"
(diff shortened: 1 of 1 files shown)


jacktasia:
Awesome! Let me know if you have any questions.

jacktasia:
@jonboiser I am a bit confused by the `[WIP]`, is this ready to merge as far as you're concerned?

jonboiser:
WIP == 'Work in progress'. I still need to write rules for variables and classes/types

jonboiser:
What is the benefit of setting `:type` in the plist? When I test out these rules,

```
foo = () => 1
foo = 1

foo()
console.log foo
```

it doesn't seem to differentiate between the types when I jump. Is there a way of telling that `foo()` should jump to `foo` (type is function)?

jacktasia:
Great question. `type` is mainly used with `dumb-jump-language-contexts`, so it would know that `foo()` should only jump to a function definition.

jonboiser:
I'm not sure why my tests are failing. I've been testing them here https://regex101.com/r/kZ6xM8/1.

I'm assuming these are perl-compatible regex?

jacktasia:
> I'm assuming these are perl-compatible regex?

They are for `ag`,  but not `grep` since not all versions of `grep` support `-P` (only `-E`).  Since most people use dumb-jump with `ag`, I would just not bother supporting `grep` if I was you. That is, switch from `:supports ("ag" "grep")` to just `:supports ("ag")` for your rules.

jonboiser:
@jacktasia I think this is ready for review. I've been testing it out on this script

https://gist.github.com/jonboiser/955fc16d8aa2803c550726396d3852bc

The language contexts thing looks really powerful, but might be left for a later day, so this PR just has basic rules.

Also, I noticed that rules in other languages still jump to commented code. I disallowed that in my rules, but had to duplicate that logic. Maybe a future feature will be to define rules for comments, and have the searcher ignore such strings?

jacktasia:
>  I think this is ready for review. I've been testing it out on this script

Cool. Thanks for doing this. I'll take a closer look tonight, but this looks pretty good.

> Maybe a future feature will be to define rules for comments, and have the searcher ignore such strings?

This is an interesting idea. In practice, I haven't run into this, but I certainly understand the appeal for certain projects. Feel free to make a new issue for it so that a discussion can get started if others agree. 

Thanks!

PR #62, opened 2016-07-14
Add extra variable matches for Rust

meqif:
Hiya!

As I kept using dumb-jump with Rust, I noticed a couple of variable matches were missing — structure fields and enumeration variants. This PR adds support for those.

Sorry for not catching all the matches in the first couple PRs. 😅


jacktasia:
Awesome, thanks!

meqif:
Thank you!

PR #65, opened 2016-07-29
Extend Ruby support

knu:
This adds support for a few more kinds of constructs to define methods in Ruby.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -205,9 +205,10 @@
            :tests ("test = 1234")
            :not ("if test == 1234"))
 
-    (:type "function" :supports ("ag" "grep") :language "ruby"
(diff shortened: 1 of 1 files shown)


jacktasia:
Thanks for the PR. It looks like one of the [rules you added is failing](https://travis-ci.org/jacktasia/dumb-jump/jobs/148289676#L561).

jacktasia:
So instead of:

 `:regex "\\bdefine(?:_singleton)?_method\\s*(?:\\(\\s*)?:JJJ\\j"` 

I think you want something like:

`:regex "\\bdefine(?:_singleton|_instance)?_method\\s*(?:\\(\\s*)?:JJJ\\b"`

PR #73, opened 2016-08-11
C/C++ support

xianbaum:
Hi! I think these regex for c/c++ should work. It should have a much higher success rate than my last PR and it should pass all the tests. Sorry I closed the last pull request (unless that's what I was supposed to do), I hard reset my fork to this branch's latest commit and it automatically closed.

The projects I tested on were Lua (a C project) and SuperTux (a C++ project) and it seems to perform well in both projects.


jacktasia:
Awesome! Thanks!

PR #74, opened 2016-08-13
Don't reinvent the xref / find-tag-marker-ring functionality

purcell:
With these changes, dumb-jump should play nicely with the standard built-in mechanisms for jumping to cross-references. (xref is new as of Emacs 25.)

(Haven't run tests locally, sorry: they hung for me after running 2/86.)

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -116,11 +116,6 @@
   "If non-nil Dumb Jump will ignore the context of point when jumping"
   :group 'dumb-jump)
 
-(defcustom dumb-jump-last-location
-  '()
-  "History of last locations when jumping"
-  :group 'dumb-jump)
-
 (defcustom dumb-jump-find-rules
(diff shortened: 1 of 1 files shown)


jacktasia:
Thanks!

dumb-jump.el
```diff
@@ -943,8 +927,9 @@ denoter file/dir is found or uses dumb-jump-default-profile"
 
 (defun dumb-jump-goto-file-line (thefile theline pos)
   "Open THEFILE and go line THELINE"
-  ;(dumb-jump-message "Jumping to file '%s' line %s" thefile theline)
-  (ring-insert find-tag-marker-ring (point-marker))
+  (if (fboundp 'xref-push-marker-stack)
+      (xref-push-marker-stack)
```
Fuco1: I just got this package installed and think, "hmm, `dumb-jump-back`, this could use `xref` pop..." I just try it for fun and it works!  I forever love you @purcell <3

purcell: Emacs conventions are a wonderful thing!

PR #86, opened 2016-09-28
add :type annotations to all defcustom variables

wbolster:
…so that the customize interface transforms the current spaghetti into pure awesomeness.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -49,82 +49,100 @@
 (defcustom dumb-jump-window
   'current
   "Which window to use when jumping valid options are 'current (default) or 'other."
-  :group 'dumb-jump)
+  :group 'dumb-jump
+  :type '(choice (const :tag "Current window" current)
+                 (const :tag "Other window" other)))
 
 (defcustom dumb-jump-selector
   'popup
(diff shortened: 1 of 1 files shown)


wbolster:
this is much better than the current huge chunk of elisp that shows up without the the `:type` annotations, don't you think? :)

image

dumb-jump.el
```diff
@@ -49,82 +49,100 @@
...
-  :group 'dumb-jump)
+  :group 'dumb-jump
+  :type 'string)
 
 (defcustom dumb-jump-fallback-search
   t
   "If nothing is found with normal search fallback to searching the fallback regex."
-  :group 'dumb-jump)
+  :group 'dumb-jump
+  :type 'string)
```
jacktasia: This should be boolean.

jacktasia:
> this is much better than the current huge chunk of elisp that shows up without the the :type annotations, don't you think? :)

I do! Thanks so much for taking the time to add these. I just saw one incorrect type, which I commented on.

dumb-jump.el
wbolster: fixed

jacktasia:
Thanks!

PR #91, opened 2016-11-17
Add haml and slim to Ruby language files

ibrahima:
I don't know if this is a good idea, since technically they aren't ruby files, but functions used in them can only be from ruby files, some of which may be helpers defined in the project. This works for me but I'm not sure if it's generally a good idea, but it seems like if js and jsx can share a language definition, perhaps so can ruby and its templating languages? Feedback appreciated!

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -666,6 +666,8 @@
     (:language "php" :ext "php" :agtype "php")
     (:language "php" :ext "inc" :agtype "php")
     (:language "ruby" :ext "rb" :agtype "ruby")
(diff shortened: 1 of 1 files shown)


jacktasia:
Thanks! I can't see any downsides to adding these. Seems super low risk with a high reward for Ruby people, so let's do it.

ibrahima:
Awesome thanks! I haven't read the entire source so wasn't sure if there were any possible problems.

jacktasia:
FYI Your change to dumb-jump is now available via MELPA. Thanks again!

Issue #92, opened 2016-11-20
dumb-jump cannot be used with files whose extensions end in numbers


jschwab:
I am in the process of trying to add fortran support to dumb-jump.

The regular expressions in `dumb-jump-parse-response-line`, i.e.

    (s-split ":?[0-9]+:" resp-line)
    (s-match ":?\\([0-9]+\\):" resp-line)

don't correctly handle the case when the filename ends in a number.  Since the colon is optional, they confuse the file extension and the line number. 
 
In this case, the value of resp-line looks like `"/path/to/file.f90:1701:      subroutine test(foo)"`

jacktasia:
> I am in the process of trying to add fortran support to dumb-jump.

Awesome!

> don't correctly handle the case when the filename ends in a number. Since the colon is optional, they confuse the file extension and the line number.

Should be fixed now. Please let me know if you run into anything else on your way.

PR #93, opened 2016-11-20
Add fortran support

jschwab:
Thanks for the quick fix of #92 !

Here's an implementation of basic fortran support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ There is currently basic support for the following languages:
 * Faust
 * R
 * Bash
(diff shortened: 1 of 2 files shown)


jacktasia:
Thanks! 🎉

jacktasia:
Oh, FYI it can take a few hours for changes to be picked up by MELPA. I'll comment again when dumb-jump is available with your changes.

jacktasia:
It's on MELPA now.

PR #94, opened 2016-11-26
Support ripgrep, GNU grep, and Windows

mwolson:
* New customization options for rg
* New customization option dumb-jump-gnu-grep-args is used when we've detected GNU grep
* Remove negative look-behind from patterns that don't need it, for rg compatibility
* Remove uses of empty match groups, backreferences, and chars that don't need escaping
* Remove dumb-jump-shell-escape in favor of shell-quote-argument, to simplify dealing with Windows
* Change a ruby test to not depend on case-insensitivity
* Escape `{` as `[{]` in some patterns for compatibility with rg
* Modify dumb-jump-ag-installed? and variants to memoize their results, so that we only need to call
  a minimal number of shell commands each time we want to use dumb-jump-go
* Run tests using stdin from a buffer, to avoid escaping problems with `echo` on Windows
* Don't use single quotes in shell commands, since Windows doesn't support that
* Quote ignored directories in command string
* dumb-jump-quick-look: Fix small docstring typo
* Introduce a `variant` argument to some function calls
* For GNU grep only, replace `\s` with `[[:space:]]`
* For ripgrep only, replace leading `-` with `[-]`
* Do not include LANG=C prefix if we're on Windows
* Do not provide exclude or include arguments to GNU grep since it doesn't support them


jacktasia:
Wow! Thanks for the PR! There's a lot going on here. Give me a bit to go over it.

mwolson:
I still have a commit pending to add some rg tests and update docs, should
land later today.

mwolson:
Should be ready for a look now

dumb-jump.el
```diff
@@ -1314,26 +1394,58 @@ Ffrom the ROOT project CONFIG-FILE."
...
+        ""
+        (dumb-jump-concat-command cmd exclude-args regex-args proj))))
+
+(defun dumb-jump-generate-rg-command (look-for cur-file proj regexes lang exclude-paths)
+  "Generate the rg response based on the needle LOOK-FOR in the directory PROJ."
+  (let* ((filled-regexes (dumb-jump-populate-regexes look-for regexes 'rg))
+         (agtypes (dumb-jump-get-ag-type-by-language lang))
+         (cmd (concat dumb-jump-rg-cmd
+                      " --color never --no-heading --line-number"
+                      (s-join "" (--map (format " --type %s" it) agtypes))))
```
jacktasia: Looks like `agtypes` can't be used with `rg` since some don't line up. For instance, `ag` uses `python` and `rg` uses `py` for `--type`, which basically means python won't work if you have `rg` installed.

mwolson: Addressed in 37be7f6

jacktasia:
Thanks!

PR #96, opened 2016-12-15
Add `dumb-jump-selector` for `helm`

deuill:
This commit adds an additional option for `dumb-jumb-selector` for using Helm to browse multi-choice jumps. A `if` clause has been changed to a `cond` to accomodate the additional rule in a more elegant manner, but the code is otherwise fairly similar to work done for `ivy` support.

Tests are forthcoming.

P.S.: Mainly opening this to test the waters, this is pretty much my first foray into Lisp so feel free to comment/berate. Not sure if the lack of `helm` suport is on purpose or if nobody did the work, but thinking it may be useful for Spacemacs users. 

P.P.S.: As mentioned above, I'll be adding tests in a subsequent commit once I figure out how to do it. ;)

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -926,7 +926,7 @@ Optionally pass t for RUN-NOT-TESTS to see a list of all failed rules"
           (dumb-jump-result-follow result))))
 
 (defun dumb-jump-prompt-user-for-choice (proj results)
-  "Put a PROJ's list of RESULTS in a 'popup-menu' (or ivy) for user to select.  Filters PROJ path from files for display."
+  "Put a PROJ's list of RESULTS in a 'popup-menu' (or helm/ivy) for user to select.  Filters PROJ path from files for display."
   (let* ((choices (-map (lambda (result)
                           (format "%s:%s %s"
                                   (s-replace proj "" (plist-get result :path))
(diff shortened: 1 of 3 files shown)


dumb-jump.el
```diff
@@ -3,6 +3,7 @@
 ;; Copyright (C) 2015-2016 jack angers
 ;; Author: jack angers
 ;; Version: 0.4.3
+;; Package-Version: 20161126.2045
```
deuill: Should probably remove this line as well.

jacktasia:
I am totally cool with `helm` support. I never added it because I quickly moved to `ivy` and no one ever requested helm support, but people would almost certainly use it. Especially if a PR was made for spacemacs so the higher level `ivy` or `helm` option was automatically supported in dumb-jump.

deuill:
Cool, I'll clean this up, add tests and open a PR against Spacemacs for setting the correct `dumb-jump-selector` based on the current completion framework.

deuill:
Formatting fixed and test pushed. Also opened a Spacemacs PR here: https://github.com/syl20bnr/spacemacs/pull/7987

The only thing potentially missing is some kind of colorization of filenames and line numbers in the completion window, like in [`helm-ag`](https://github.com/syohex/emacs-helm-ag/blob/997107a53abd3bda06d52e7021c87527c5747389/helm-ag.el#L404), I didn't want to pollute the source code with what is essentially a nice-to-have though.

jacktasia:
Thanks!

Issue #36, opened 2016-05-27
Manually set project root


cweiske:
dumb-jump uses the first project root indicator file it finds to set the project root to its containing directory.

This approach does not work for complex projects in which multiple git checkouts are used:

```
php-typo3-project/
+ typo3conf/
|  + ext/
|     + extension/
|        + .gitignore
|     + extension2/
|        + .gitignore
+ typo3_src/
+ .gitignore
```

When using `dumb-jump` on a file in `php-typo3-project/typo3conf/ext/extension/src/Foo.php`, it finds `php-typo3-project/typo3conf/ext/extension/` as project root.
I'd rather want to have `php-typo3-project` as root.

jacktasia:
Thanks for opening this. This is a great point. I will take a look at this over the weekend.

I think the easiest solution would be if we had a file called like `.dumbjumpignore` that was put in every directory that could mistakenly be treated as the project root. Any thoughts or concerns with this?

cweiske:
Thank you! I'll wait for the next melpa release to try it out.

PR #103, opened 2017-03-12
Project root detection: Ignore sub-folders using '.dumbjumpignore' (#36)

netromdk:
If a project has a ".dumbjumpignore" file it won't be detected as a project root even though it has
another project root denoter, like ".git" etc. This makes it possible to have a fallback top project
so searches can lookup things from other sub-projects.

Added a test case to verify it.

It fixes issue #36.


jacktasia:
Awesome, thanks!

netromdk:
Would it be an idea to extend the README.md about the ".dumbjumpignore" file and its usage? Thanks.

jacktasia:
Agreed. Thanks for the reminder. I'll do this later today.

Issue #104, opened 2017-03-13
Implement jumping to matches not in current file


netromdk:
In some cases, you want to jump to an external definition, but there are local uses in the current file that you don't want it to match at all.

Perhaps adding `dumb-jump-go-non-current` and `dumb-jump-go-non-current-other-window` interactive functions for this purpose?

Thanks

netromdk:
Or it could sort matches for current file last, so it still jumps locally if no external matches are found.

jacktasia:
I agree that this is a good idea. I think I like `dumb-jump-go-prefer-external` and `dumb-jump-go-prefer-external-other-window` best.

PR #106, opened 2017-03-13
Jump and prefer external + other window (#104)

netromdk:
`dumb-jump-go-prefer-external` works like `dumb-jump-go` except it will sort external matches before the current file. `dumb-jump-go-prefer-external-other-window` will find file in other window by preferring external files, too.

Added a test case: dumb-jump-prefer-external

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
 language: generic
 before_install:
   - sudo add-apt-repository ppa:gekkio/ag -y
+  - sudo add-apt-repository ppa:git-core/ppa -y
   - sudo apt-get -qq update
-  - sudo apt-get install -y silversearcher-ag
+  - sudo apt-get install -y silversearcher-ag git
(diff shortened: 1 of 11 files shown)


jacktasia:
Oh man, sorry I somehow missed this PR completely. I'll review it after work tonight.

jacktasia:
Thanks for the PR! I think this is a really good idea. I've found a small issue. When using `dumb-jump-go-prefer-external` if there aremore than two candidates in the same file it won't force itself to use the external file.

For example if you add this to the top of `external.cpp`:

```
int ExternalThing {
  return 0;
};
``` 

so it looks like:

```
#include "external.h"

int ExternalThing {
  return 0;
};

class ExternalThing {
};

int main(int argc, char **argv) {
  ExternalThing et;
}
```

and jump from the last `ExternalThing` you'll see what I mean. It seems to even occur if add it in commented out.

netromdk:
@jacktasia finally :D

Issue #34, opened 2016-05-27
Use git grep when available


catern:
git grep should be faster than both ag and grep, when it is available.

jacktasia:
Thanks for opening this. This is a really good point. I will definitely look into this.

netromdk:
@jacktasia this issue should be closed now, I guess.

PR #107, opened 2017-03-15
Implemented support for git grep (#34)

netromdk:
Can be used by setting `(setq dumb-jump-searcher "git-grep")`.

Also added test cases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: generic
 before_install:
   - sudo add-apt-repository ppa:gekkio/ag -y
   - sudo apt-get -qq update
-  - sudo apt-get install -y silversearcher-ag
+  - sudo apt-get install -y silversearcher-ag git-all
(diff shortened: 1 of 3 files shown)


netromdk:
The coverage decreased because travis doesn't have git grep installed when running the tests.

jacktasia:
Wow! Thanks for the PR!

> The coverage decreased because travis doesn't have git grep installed when running the tests.

I think if we add `git-all` to the [end of this line](https://github.com/jacktasia/dumb-jump/blob/master/.travis.yml#L5) so it's `sudo apt-get install -y silversearcher-ag git-all` instead it might work.

jacktasia:
Hmm, TravisCI didn't seem to like `git-all` I guess we should just try `git` instead.

netromdk:
Totally agree :)

netromdk:
@jacktasia finally! Hehe.

jacktasia:
Awesome! Thanks!

PR #108, opened 2017-03-16
Only search file extensions with git-grep

netromdk:
`ag` and `rg` only searches the file extensions by the given language and now so does `git grep`. It's a lot faster, too! Even when compared to `ag`.

I have pretty big work repo and I tired using `ag` and `git grep` on the same symbol:

- `ag` took 1.381814956665039 seconds
- `git grep` w/o using extensions took 10.966195106506348 seconds
- **`git grep` w.  extensions took 0.087921142578125 seconds**


jacktasia:
Wow! 💨  Thanks!

netromdk:
Great. You're welcome. :)

Issue #109, opened 2017-03-16
Possibility of having git-grep in git repos and fallback to ag


netromdk:
I was thinking that people would probably prefer to use `git grep` in a git repo because it's usually faster, and now they can with commit https://github.com/jacktasia/dumb-jump/commit/a90ae445a2d9cb3feeea3ce7752a9d723838e986.

However, it's not always useful to have `(setq dumb-jump-searcher "git-grep")` if you sometimes search non-git projects. In this case, I was wondering if it would be helpful with an easy way of defining a fallback searcher? In my ideal scenario, it will pick `git grep` in a git repo and otherwise use `dumb-jump-searcher`. Thoughts?

netromdk:
Perhaps there could be a `(setq dumb-jump-use-git-grep-in-git-repo t)` or something, and `dumb-jump-pick-grep-variant` would then use `git grep` if the project has a ".git" denoter and `git grep` is installed.

jacktasia:
I haven't fully thought this out yet but I was kind of thinking of having `dumb-jump-searcher`* default to `nil` and then when `nil` it will pick the best option which seems to be (in order): `git-grep` (if it's a git project), `ag`, `rg`, `grep`. If it's anything but `git-grep` it would still try to use `git grep` if it was a git project...but then also add a new config called `dumb-jump-force-searcher` that would also default to `nil` but if it was set it would just blindly use whatever was set.

So

### Example 1
```
(setq dumb-jump-searcher nil)
(setq dumb-jump-force-searcher nil)
```
would use `git-grep` if a git repo and if not then use whatever was available in order `ag`, `rg`, grep.

### Example 2
```
(setq dumb-jump-searcher nil) 
(setq dumb-jump-force-searcher "ag")
```
would always use `ag` and `dumb-jump-searcher` would effectively be ignored (even if it was set)

*It'd probably be a good idea to drop `dumb-jump-searcher` and call it `dump-jump-prefer-searcher` to make it more clear.

What I like about this is it would have a sane default "out of the box" that is the fastset and people wouldn't have to worry about what it's doing under the hood, but then if they really care they can force the searcher of your choice.

Thoughts? Any flaws with this thinking?

netromdk:
I agree with your two variables, making it possible to force a searcher if necessary (like wanting `rg` which will not be chosen if both `git grep` and `ag` are installed, for instance), and still being able to define a preferable searcher. 

[...]

PR #111, opened 2017-03-18
Forced, preferred, or optimal searcher selection

netromdk:
In relation to #109, two new variables have been created:
- `dumb-jump-prefer-searcher`
- `dumb-jump-force-searcher`

Both default to `nil`.

`dumb-jump-prefer-searcher` can use `'ag`, `'rg`, `'git-grep`, `'gnu-grep`, or `'grep`. If `nil` then the most optimal searcher will be chosen at runtime.

`dumb-jump-force-searcher` forces to use searcher `'ag`, `'rg`, `'git-grep`, `'gnu-grep`, or `'grep`. Set to `nil` to not force anything and use `dumb-jump-prefer-searcher` or most optimal searcher.

In addition, if no searcher is forced, project root has a `.git`, and `git-grep` is installed then it will use that instead. Otherwise, it uses the fastest searcher that is installed in the following order: `ag`, `rg`, `gnu-grep`, or `grep`.


netromdk:
Awesome! Almost 100% coverage now, @jacktasia.

jacktasia:
This is great! Thanks!

PR #115, opened 2017-03-21
Helm preview choices on C-j

netromdk:
When using helm it is usually the case that <kbd>C-j</kbd> executes a persistent action, meaning to preview something while keeping the helm candidates buffer open.

I've implemented so <kbd>C-j</kbd> will preview the match at current line of the helm candidates lines in another temporary buffer until (1) hitting <kbd>enter</kbd> and it jumps to the file at that line, or (2) <kbd>C-g</kbd> will abort helm and return to the original buffer position.


jacktasia:
Thanks!

PR #116, opened 2017-03-22
Git grep untracked

netromdk:
Sometimes you're working on files that are untracked by the git repo but you still want to be able to search in them.

The following enables that:
`(setq dumb-jump-git-grep-search-untracked t)`

It defaults to `nil`, i.e. unused.


jacktasia:
Thanks for the PR!  I kind of think this should default to `t`. My thinking is this is more in line with our philosophy of "just working" over anything else. I'd rather it be slightly slower and "just work" then _seem_ broken and have the user need to check the docs/code to see why it isn't working as expected. Thoughts/concerns?

netromdk:
That's a good point. I couldn't decide whether to enable it by default or not. I will enable it.

jacktasia:
Awesome 🎉  Thanks!

PR #117, opened 2017-03-24
Improve PHP variable and class detection

cweiske:
I don't have a test for the "space with keyword" because you only use "test" as keyword. But when jumping to definition, emacs also includes a leading dollar sign into the word, thus searches for `$foo` - so the test would need to look for `$test` to make sense.

Same for `::$foo`

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -449,8 +449,8 @@ using searcher git-grep."
            :tests ("function test()" "function test ()"))
 
     (:type "variable" :supports ("ag" "grep" "rg" "git-grep") :language "php"
-           :regex "JJJ\\s*=\\s*"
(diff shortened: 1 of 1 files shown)


dumb-jump.el
```diff
@@ -449,8 +449,8 @@ using searcher git-grep."
            :tests ("function test()" "function test ()"))
 
     (:type "variable" :supports ("ag" "grep" "rg" "git-grep") :language "php"
-           :regex "JJJ\\s*=\\s*"
-           :tests ("$test = 1234"))
+           :regex "( |->|\\$|::)JJJ\\s*=\\s*"
```
netromdk: You should instead use `\\s` than using a "space".

jacktasia:
Thanks!

PR #119, opened 2017-03-26
Sort matches better and correctly filter current file

netromdk:
Two things:

1. The current file wasn't correctly being determined in all cases. Let's say the project root is "/some/path" and the `cur-file` is "/some/path/Controller.cc". However, some results are relative to the file the searcher was initiated from, so in this case, it would be "Controller.cc" and not being matched. Another case is for results of files in any arbitrary subfolder. For instance, if `cur-file` is "/some/path/src/db/Controller.cc" the relative path is "src/db/Controller.cc", which is neither of `cur-file` or "Controller.cc" - but it is handled by using `default-directory`.

2. For sanity and convenience, I'm now sorting non-current files by shortest path/lowest line number first. This means that the chance of having a result sorted high to the top is higher for matches in files close to the current file, i.e. folder names and "../../.." etc. makes the path longer. So instead of seeing:

- Controller.cc:101    static void loadQtResources()
- ../monitor/ui/ConfigEditor.cc:27    static void loadQtResources()
- OtherStuff.cc:99    static void loadQtResources()
- ../updatemonitor/UpdateMonitor.cc:20    static void loadQtResources()

It would show:
- Controller.cc:101    static void loadQtResources()
- OtherStuff.cc:99    static void loadQtResources()
- ../monitor/ui/ConfigEditor.cc:27    static void loadQtResources()
- ../updatemonitor/UpdateMonitor.cc:20    static void loadQtResources()

With "Controller.cc" being the current file so it is sorted first, of course.

I hope this makes sense.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -1350,28 +1350,42 @@ PREFER-EXTERNAL will sort current file last."
          (match-sorted (-sort (lambda (x y) (< (plist-get x :diff) (plist-get y :diff))) results))
          (match-no-comments (dumb-jump-filter-no-start-comments match-sorted lang))
 
+         ;; Find the relative current file path by the project root. In some cases the results will
+         ;; not be absolute but relative and the "current file" filters must match in both cases.
+         (rel-cur-file
+          (if (s-starts-with? proj-root cur-file)
+              (substring cur-file (1+ (length proj-root)) (length cur-file))
+            cur-file))
+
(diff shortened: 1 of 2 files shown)


netromdk:
Any update regarding this? 😊

jacktasia:
> Any update regarding this, @jacktasia ? 😊

Sorry for the delay. Thanks!

netromdk:
No worries. Thanks for merging!

PR #123, opened 2017-03-28
Introduced aggressive mode

### LINKED ISSUE #122: Jumping vs. showing candidates
In `dumb-jump-handle-results` there's a `do-var-jump` that I find a little flawed. Especially the `(string= ctx-type "")` part from my excerpt beneath:

```elisp
(matches
 (if (not prefer-external)
     (-distinct
      (append (dumb-jump-current-file-results cur-file match-cur-file-front)
              (dumb-jump-current-file-results rel-cur-file match-cur-file-front)))
   match-cur-file-front))

(var-to-jump (car matches))

(do-var-jump (and (or (= (length matches) 1)
                      (string= ctx-type "variable")
                      (string= ctx-type ""))
                  var-to-jump))
```

Before I fixed the absolute/relative current file matching in https://github.com/jacktasia/dumb-jump/commit/e23bff8b918c41dd8dc9a13cc117e919bd21e58a it would not trigger `do-var-jump` because `matches` wouldn't get a size of 1 often. However, now, if there is one match or the `ctx-type` is `""` then it just jumps. In most situations I only want it to show me the candidates to select from - if there is only one candidate at all, then it's fine that it just jumps.

I think you jump when context is empty due to some of the tests perhaps?
Hope it makes sense.

netromdk:
This PR fixes #122 which has more information about the thought processes.

Essentially, there's now a new variable `dumb-jump-aggressive` that defaults to `t` and makes Dumb Jump work like it normally does. If it is set to `nil` then it will only do a variable jump if there's a single match; in the current file and any other files.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -827,6 +827,13 @@ a symbol then it's probably a function call"
   :group 'dumb-jump
   :type 'hook)
 
+(defcustom dumb-jump-aggressive
+  t
+  "If `t' jump aggressively with the possiblity of a false
+positive. If `nil' always show list of more than 1 match."
+  :group 'dumb-jump
+  :type 'boolean)
+
 (defun dumb-jump-message-prin1 (str &rest args)
   "Helper function when debugging apply STR 'prin1-to-string' to all ARGS."
   (apply 'message str (-map 'prin1-to-string args)))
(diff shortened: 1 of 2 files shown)


jacktasia:
Thanks!

PR #125, opened 2017-03-30
Shell support

### LINKED ISSUE #124: Independent of underlying shell?
I find that when I've enabled Iterm2's shell integration (https://www.iterm2.com/documentation-shell-integration.html), `dumb-jump` starts finding filenames like 

```
~/dir/dir/dir/^[]1337;RemoteHost=foobar@foobar-MacBook-Pro-2.local^G^[]1337;CurrentDir=/Users/foorbar/dir/dir^G^[]1337;ShellIntegrationVersion=5;shell=zsh^G/Users/foobar/dir/dir/dir/filename.rb
```

Wouldn't it be better if the dumb-jump commands ran in a shell that wasn't affected by these sorts of login customizations?

netromdk:
Fixes #124.

No profile or RC files are loaded when executing a search to speed things up and avoid unnecessary things being run, which can potentially yield erroneous path names.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -1570,10 +1570,10 @@ searcher symbol."
       "-icf")
 
      ((string-equal "bash" base-name)
-      "--noprofile --norc -c")
+      "-c")
 
      (t
-      "-c"))))
+      shell-command-switch))))
 
 ;; TODO: rename dumb-jump-run-definition-command
 (defun dumb-jump-run-command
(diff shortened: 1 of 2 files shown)


jacktasia:
Thanks for the PR! 

I am pretty sure the tests are failing because (at least for `bash`) the `rc` and `profile` are required to ensure that the searchers are in the `PATH`.

netromdk:
You're welcome. There we go!

jacktasia:
Thanks!

PR #126, opened 2017-04-11
No longer provide a feature in test-helper.el

tarsius:
The file isn't a library intended to be loaded with `require`.
Instead `ert-runner` loads it using `load`.  Other packages that
use `ert-runner` also come with a file named test-helper.el and
unfortunately many of those also provide `test-helper`, which
leads to conflicts.

### DIFF
--- a/test/test-helper.el
+++ b/test/test-helper.el
@@ -11,5 +11,4 @@
 (eval-when-compile
   (require 'cl))
 
-(provide 'test-helper)
 ;;; test-helper.el ends here


jacktasia:
Thanks for the PR.

> Other packages that use ert-runner also come with a file named test-helper.el and
unfortunately many of those also provide test-helper, which leads to conflicts.

In practice, I don't understand where these conflicts are actually happening if when a package is installed via MELPA the test code is never even downloaded on a user's machine.

tarsius:
It can happen when a a package is *not* installed via melpa. I for example use my [`borg`](https://github.com/emacscollective/borg) package manager (which doesn't have an issue with this. Other alternative package managers might though. "Doing it by hand" too.

jacktasia:
> It can happen when a package is not installed via melpa. I for example use my borg package manager, which doesn't have an issue with this. Other alternative package managers might though. "Doing it by hand" too.

Ah, ok. Thanks.

tarsius:
Thanks to you too!

PR #143, opened 2017-06-08
change aggressive default to nil

jacktasia:


jacktasia:
@netromdk (or anyone else reading) Can you think of any reason not to merge this?

netromdk:
@jacktasia I personally use `nil` and I think it's the "best" default because the user is still presented with the choices.

PR #146, opened 2017-06-27
add css to javascript language

redguardtoo:
it's common to reference css class name in javascript/html/jsx. For example, if jQuery, we use css class name as dom selector. In JSX , the css name could be used as a javascript object (using css-modules)


jacktasia:
Thanks!

PR #149, opened 2017-07-21
Add basic ocaml support + lexical scoping

rgrinberg:
Will hopefully keep improving this as I use this


jacktasia:
Thanks for the PR! The only reason I can think for why it's only failing on emacs 24.3 in Travis is because of the lexical-bindings change. Thoughts?

rgrinberg:
I think that lexical bindings are a very good thing to use, but emacs 24.3 isn't that old and I'm not really understanding that error. So I'll get rid of it until you decide to drop support for 24.3

jacktasia:
> I think that lexical bindings are a very good thing to use, but emacs 24.3 isn't that old and I'm not really understanding that error. So I'll get rid of it until you decide to drop support for 24.3

Thanks! I totally agree.

PR #152, opened 2017-07-25
#150 enable lexical bindings

### LINKED ISSUE #150: lexical-bindings and emacs 24.3
Using lexical-bindings [seems to break tests for 24.3](https://travis-ci.org/jacktasia/dumb-jump/jobs/255939067) as found via #149

akhramov:
It's been observed that tests fail on GNU Emacs 24.3 if lexical-bindings are
enabled. There could be various reasons for this, but most obvious is
that Emacs 24.3 had `byte-compile-add-to-list` function, which was used
as byte-compiled version of `add-to-list`:
https://github.com/emacs-mirror/emacs/commit/0b31660d3c10a0f8e243dd67bd0ecaf2c847d1e6#diff-38337c5e732c14fa599af4a0f6e53f18L4272

ert-runner compiles files before the test run, so tests used the
compiled version.

This Pull Request rewrites functions that depended on add-to-list, and hopefully fixes #150.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ If you want to stop a directory from registering as the project root (and have D
 
 * `(setq dumb-jump-default-project "~/code")` to change default project if one is not found (defaults to `~`)
 * `(setq dumb-jump-quiet t)` if Dumb Jump is too chatty.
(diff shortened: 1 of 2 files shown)


jacktasia:
Awesome! Thanks!

Issue #156, opened 2017-08-12
Doesn't recognize extension-less Perl files using `cperl-mode`


genehack:
When editing a Perl file that doesn't have an extension (e.g., a command script in Perl), using `cperl-mode`, currently `dumb-jump` reports `Could not find rules for ’. file’.`.

Since `cperl-mode` is the most common mode for editing Perl files in Emacs, this is unfortunate. Pull request to fix this incoming in just a second tho. 8^)

PR #157, opened 2017-08-12
Fix `dumb-jump-get-language-from-mode` for `cperl-mode` [#156]

genehack:
The most commonly used mode for editing Perl files is `cperl-mode`. Prior to this patch, opening a Perl file without an extension would give an error message; this maps `cperl-mode` to `perl` as expected and then `dumb-jump` works.


genehack:
This fixes #156 for me.

jacktasia:
Thanks!

PR #160, opened 2017-09-06
Update rules for ruby

knu:
- Add support for mass assignment
- Make rules stricter to reduce noises
- Use `\W` intead of `\j`, since the latter does not seem to work well (namely with git-grep)

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -423,28 +423,28 @@ or most optimal searcher."
            :tests ("class test(object):" "class test:")
            :not ("class testnot:" "class testnot(object):"))
 
-    ;; ruby
-    (:type "variable" :supports ("ag" "grep" "rg" "git-grep") :language "ruby"
-           :regex "\\s*JJJ\\s*=[^=\\n]+"
-           :tests ("test = 1234")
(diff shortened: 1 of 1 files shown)


knu:
Another point: This adds support for git-grep, while removing use of (rather pointless) `(?:...)` which git-grep with `-E` given does not understand.

I think you could just use `-P` because almost all distributions of git has PCRE support compiled in including those of macOS and FreeBSD, not just Linux.  However, that is another story.

knu:
Will fix the build failures.

jacktasia:
Thanks for working on this! I'll merge it as soon as all the tests pass. Please let me know if I can help.

> I think you could just use -P because almost all distributions of git has PCRE support compiled in including those of macOS and FreeBSD, not just Linux. 

I can't remember now, but there was some OS/Distro back when I started this project where the default/built-in `grep` did not have support for `-P`. It _might_ have been an older version macOS. That said, I am open to switching to using it I'd just need to do some tests and explicitly check if it `-P` is supported and then warn the user if it won't work.

jacktasia:
Thanks!

Issue #159, opened 2017-09-04
Julia-lang support


AShedko:
Is there a chance support for julia (julialang.org) is coming?

It just reached 46 on TIOBE this august and it has an emacs mode.

jacktasia:
@AShedko Thanks for opening this issue! I've just pushed up _basic_ julia support. I have zero experience with julia, but I was able to successfully jump around[ this julia repo I found](https://github.com/GiovineItalia/Gadfly.jl).

The changes should be available via MELPA in ~6  hours. Please let me know if you run into any issues with it or have any specific requests. Thanks!

jacktasia:
This is now on MELPA. Please let me know if you run into any issues with it or have any specific requests. Thanks!

PR #162, opened 2017-09-10
Jl support

AShedko:
Some changes
* made types searchable
* added some function decorators

response for this
https://github.com/jacktasia/dumb-jump/issues/159

SIde note:
Does spacemacs lag behind melpa, as the changes are not awailable there?

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -653,6 +653,25 @@ or most optimal searcher."
            :regex "^\\s*module\\s*type\\s*\\bJJJ\\b"
            :tests ("module type test ="))
 
+    ;; julia
+    (:type "variable" :supports ("ag" "grep" "rg" "git-grep") :language "julia"
+           :regex "\\s*JJJ\\s*=[^=\\n]+" :tests ("test = 1234") :not ("if test == 1234:"))
+
+    (:type "function" :supports ("ag" "grep" "rg" "git-grep") :language "julia"
+           :regex "(?:\s*|@inline|@noinline)function\\s*JJJ\\b\\s*\\\("
+           :tests ("\tfunction test(asdf)" "function test()")
+           :not ("\tfunction testnot(asdf)" "function testnot()"))
+
+    (:type "type" :supports ("ag" "grep" "rg" "git-grep") :language "julia"
+           :regex "struct\\s*JJJ\\b\\s*"
+           :tests ("struct test" "struct test")
+           :not ("struct testnot" "struct testnot"))
+
+    (:type "type" :supports ("ag" "grep" "rg" "git-grep") :language "julia"
+           :regex "struct\\s*JJJ\\b\\s*"
+           :tests ("struct test" "struct test")
+           :not ("struct testnot" "struct testnot"))
+           )
     ;; lua
     (:type "variable" :supports ("ag" "grep" "rg" "git-grep") :language "lua"
            :regex "\\s*\\bJJJ\\s*=[^=\\n]+" :tests ("test = 1234") :not ("if test === 1234"))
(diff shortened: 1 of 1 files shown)


jacktasia:
Awesome, thanks!

> Does spacemacs lag behind melpa, as the changes are not available there?

I honestly don't use spacemacs, so I am not sure, but it could be:

1. There is a spacemacs julia layer and it doesn't load dumb-jump for its keybindings?
2. They use a mirror of MELPA that runs behind?

or maybe something else. if you can confirm the correct version of dumb-jump is installed _after_ you upgrade (the only way I know to check is: <kbd>M-x</kbd> `package-list-packages-no-fetch`) then it's potentially number 1 and we may want to consider creating an  issue with spacemacs.

AShedko:
Apparently they use version 2017.08.21.13 .

jacktasia:
> Apparently they use version 2017.08.21.13 .

[It doesn't look like they pin versions yet]( https://github.com/syl20bnr/spacemacs/issues/2057). Have you tried upgrading `dumb-jump` manually? Try <kbd>M-x</kbd> `package-refresh-contents` then `package-list-packages` the on that buffer <kbd>U</kbd> (shift+u) then <kbd>x</kbd>

AShedko:
I got the latest version. Thank you. unfortunately i had no luck actually using the module as i keep getting
`julia declaration not found` errors.

jacktasia:
>  unfortunately i had no luck actually using the module as i keep getting `julia declaration not found` errors.

Sorry, it didn't work. Could you please provide more information about your setup? 

- OS
- shell
- searcher (ag, grep, etc), 
- debug output (setting `(setq dumb-jump-debug t)` and then trying a jump and checking `*Messages*`)
- example of code you're jumping from/to
- check if another searcher works (`(setq dumb-jump-force-searcher 'ag)` See README)

This will help me track down where it's failing for you.

AShedko:
OS: GNU/Linux (Fedora 25) [Linux localhost.localdomain 4.12.8-200.fc25.x86_64 #1 SMP Thu Aug 17 15:39:27 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux]
shell: zsh
searcher: ag
debug output:
`

cmd: 
	ag --nocolor --nogroup --julia --jl \(\@noinline\|\@inline\)\?\\s\*function\\s\*subplot_data\\s\*\\\(\|const\\s\+subplot_data\\b\|\(mutable\)\?\\s\*struct\\s\*subplot_data\|\(type\|immutable\|abstract\)\\s\*subplot_data /home/ashedko/.julia/v0.6/Gadfly
raw results: 
	ag: unrecognized option '--jl'
Usage: ag [FILE-TYPE] [OPTIONS] PATTERN [PATH]
...

-----
DUMB JUMP DEBUG ‘dumb-jump-run-command‘ END
-----
`
setting `grep` as the searcher works.

jacktasia:
The `ag: unrecognized option '--jl'` is telling us the problem. 

My bad for not catching this in the PR, but you overrode the [existing](https://github.com/jacktasia/dumb-jump/blob/master/dumb-jump.el#L864) `jl` extension rule with bad [rg](https://github.com/BurntSushi/ripgrep/blob/master/ignore/src/types.rs#L149) and [ag](https://github.com/ggreer/the_silver_searcher/blob/master/tests/list_file_types.t#L141) types.

Fixed in https://github.com/jacktasia/dumb-jump/commit/5aa56feecfe38971d9ede687eebf6346dc85758a

Thanks for getting me the debug details so fast.

AShedko:
Glad to help, I think I`ll add some more regexes for function declaration syntax.

PR #164, opened 2017-09-11
Jl support

AShedko:
Found some more cases where search does not work.
Silly me, I forgot about functions with type templates.


jacktasia:
Thanks!

PR #165, opened 2017-09-18
add Erlang support

tsloughter:
Variable matching misses a lot of ways it is typical to bind variable in Erlang, but I don't think they could be covered without just finding everything that contains the term that is being searched for :(.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ There is currently basic support for the following languages:
 * Coq
 * Elixir
 * Emacs Lisp
+* Erlang
 * Faust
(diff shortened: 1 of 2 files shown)


jacktasia:
Awesome, thanks! 🎉 

There was just one tiny issue with one of your regex when it was [used with ripgrep](https://github.com/BurntSushi/ripgrep/issues/102), which you wouldn't have seen without ripgrep installed. Fixed here: 4988674

tsloughter:
Great, thanks!

PR #166, opened 2017-10-13
Modify format of choices; Move ivy selection function to user option

jtamagnan:
These changes are meant to enable the creation of https://github.com/jtamagnan/counsel-dumb-jump
Which is a package that enables font-lock in the ivy completion buffer for dumb-search and also makes it so that the corresponding ivy-occur buffer is fully featured, as it would be if using `counsel-rg` or `swiper`. I would be willing to bring `counsel-dumb-jump` into `dumb-jump` but I thought that as this would add some dependencies it would be best for it to be its own seperate repo. I would love your opinion on this. 

 * dumb-jump.el: Modify format of choices to match emacs grep format.
   Create dumb-jump-ivy-jump-to-selected-function variable to allow
   users to customize how ivy should select a match

 * test/dumb-jump-test.el: Fix tests to match new format

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -66,6 +66,10 @@
                  (const :tag "Helm" helm)
                  (const :tag "Ivy" ivy)))
 
+(defcustom dumb-jump-ivy-jump-to-selected-function
+  #'dumb-jump-ivy-jump-to-selected
+  "Prompts user for a choice using ivy then dumb-jump to that choice")
+
 (defcustom dumb-jump-prefer-searcher
   nil
   "The preferred searcher to use 'ag, 'rg, 'git-grep, 'gnu-grep,or 'grep.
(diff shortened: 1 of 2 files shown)


jacktasia:
Thanks for the PR! This sounds great. I'll give it a closer look tonight after work then merge and/or offer feedback. Thanks again!

jacktasia:
>  I would be willing to bring counsel-dumb-jump into dumb-jump but I thought that as this would add some dependencies it would be best for it to be its own separate repo. I would love your opinion on this.

I agree with your thinking as far as not adding any more dependencies, but I am certainly open to the idea of embedding it inside dumb-jump if you think there's a good argument for it.

Assuming you want to keep it a separate package it'd be cool if you wanted to add an `Extensions` section to the dumb-jump README ([like projectile's](https://github.com/bbatsov/projectile#extensions)) with a blurb/link about `counsel-dumb-jump`, but obviously up to you.

Thanks again!

PR #169, opened 2017-11-03
Add SCSS support

thiderman:
This might not cover the entire spec, but it covers my use cases!

Also, I couldn't run the tests because cask couldn't install. They are simple and from my understanding of the other definitions they should be at least mostly okay. :sweat_smile:

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -856,16 +856,16 @@ or most optimal searcher."
 
     ;; scss
     (:type "function" :supports ("ag" "grep" "rg" "git-grep") :language "scss"
-           :regex "@mixin JJJ\\b\\s*\\\("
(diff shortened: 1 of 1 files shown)


jacktasia:
Thanks for the PR! Looks like the tests are not passing though. Looks like you're setting the `test` value to what to jump _from_ instead of an example/test match of the actual function to jump _to_.

I was able to fix them pretty easily: https://github.com/jacktasia/dumb-jump/compare/scss-tests -> https://travis-ci.org/jacktasia/dumb-jump/jobs/296997840 if you want to update your branch so you get credit for your work...or if you don't care I can just merge my branch. Up to you. Thanks again!

jacktasia:
Looks like I could push to your branch. Thanks!

PR #175, opened 2017-11-19
fix js function call false positive

bounceme:
`.then(function(){`

Not sure about the elisp regex tho


jacktasia:
Thanks for the PR! I feel pretty good about merging this since all the tests still pass. Is the `:not` test I added a good representative of the false positive? Since I've committed I realize that it doesn't pass with the old regex when it's moved to `:tests` so I am guessing not. What would it be? Thanks!

bounceme:
yea, the `not test` shows the case where it was failing. afterwards goes straight to `pinkie/index.js` and the prototype declaration!

PR #177, opened 2017-11-23
Fixes overly lax matching of some Ruby identifiers

hinrik:
Any names with the same prefix were being considered equivalent:

```ruby
class Foo::Bar
end

class Foo::Bar::Baz < Foo::Bar
end

class Foo::Bar::Quux < Foo::Bar
```

When trying to jump to the definition of `Foo::Bar`, it would consider
all of the class definitions above to be equally good matches, and
would present a prompt instead of jumping to the right definition.


jacktasia:
Thanks!

PR #178, opened 2017-11-26
Use PCRE with GNU Grep and git-grep

hinrik:
Almost all the patterns used by dumb-jump contain non-POSIX constructs
like `[\w]`, which work in `ag` and `rg` but not in `grep`/`git grep` unless
the `-P` option is used.

This was causing parts patterns to be ignored (e.g. the final parts of
most Ruby/Crystal patterns), therefore matching too many candidates
instead of finding only the correct one.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -2056,7 +2056,7 @@ searcher symbol."
                       (if dumb-jump-git-grep-search-untracked
                           " --untracked"
                         "")
-                      " -E"))
+                      " -P"))
          (fileexps (s-join " " (--map (shell-quote-argument (format "%s/*.%s" proj it)) ggtypes)))
          (exclude-args (s-join " "
                                (--map (shell-quote-argument (concat ":(exclude)" it))
@@ -2088,7 +2088,8 @@ searcher symbol."
          (cmd (concat (if (eq system-type 'windows-nt) "" (concat dumb-jump-grep-prefix " "))
                       (if (s-ends-with? ".gz" cur-file)
                           dumb-jump-zgrep-cmd
-                        dumb-jump-grep-cmd)))
+                        dumb-jump-grep-cmd)
+                      " -P"))
          ;; TODO: GNU grep doesn't support these, so skip them
          (exclude-args "")
          (include-args "")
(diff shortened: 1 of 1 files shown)


jacktasia:
Thanks for the PR! As far as GNU grep this has [come up before](https://github.com/jacktasia/dumb-jump/pull/160#issuecomment-327551944) and I am still concerned about backwards compatibility. I think most people are using the more advanced searchers (ag/rg/etc) and for those stuck on grep it's hopefully just a degraded experience instead of potentially not working at all.

That said, part of me feels like supporting GNU grep is important because it's on almost every system and dumb-jump has a "just work" philosophy...and part of me feels like maybe support for grep should be deprecated entirely because of that philosophy since installing a more advanced searcher ensures a better experience. I am interested to hear the thoughts on this from anyone that might be reading this

Anyway, I'd like to keep this open for more discussion for GNU grep. It might make sense to split this into two PRs one for GNU grep and one for git-grep. Thoughts? I personally haven't run into this issue with git-grep. Have you @netromdk? Then my only concern is the tests. I haven't had a chance to look into why exactly they're failing again.

Thanks again for all your PRs!

netromdk:
Personally, I would rather we used all POSIX compliant regex if possible. But PCRE's fine, too, but it has to be decided what flavor is used in general. The Silver Searcher uses PCRE by default and ripgrep [seams to be support PCRE](https://doc.rust-lang.org/regex/regex/index.html#syntax). That means all the popular programs support PCRE and it would probably make sense to make PCRE the default.

However, the standard BSD grep on macOS doesn't support "-P". It's compatible with GNU grep but not in this regard. Should BSD grep be supported?

The test failures are, for the most part, about git-grep tests not being updated to match for "-P" instead of "-E".

netromdk:
@jacktasia what do you want to do with this one?

netromdk:
It should be tested if #188 fixed this.

jacktasia:
@netromdk I definitely _think_ #188 will fix this:

> This was causing parts patterns to be ignored (e.g. the final parts of
most Ruby/Crystal patterns), therefore matching too many candidates
instead of finding only the correct one.

since it sounds very similar to #186 but @hinrik If you get a chance to confirm or deny this. It'd be great. It probably won't get picked up by MELPA for a few more hours (from this writing). Thanks!

jacktasia:
My understanding is #188 solves the problem that this PR was addressing so I am going to close it. Please let me know if I am wrong.

netromdk:
There were also removals in the ruby and crystal rules. Do they need to go in, @hinrik? Perhaps you could push only those changes then.

PR #102, opened 2017-03-10
Fixed C++ function regexp for template arguments (#87)

netromdk:
I fixed C++ function regexp for template arguments and added a few C++ tests.

Test file "test/data/proj1/src/cpp/issue-87.cpp" verifies the fix of issue #87.

"Ran 106 tests, 106 results as expected (2017-03-10 23:06:41+0100)"

PR #180, opened 2017-12-02
Initial attempts to add more haskell rules

volhovm:
Alright, so I want to extend haskell support a little bit at least. This PR is currently WIP, because there are things I'm stuck with and I'd like somebody to help me (maybe). The main one is: does `ag` support multiline matching? I see a proof (here https://github.com/ggreer/the_silver_searcher/issues/682) that it does, but for some reasons my tests show otherwise.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -678,12 +678,12 @@ or most optimal searcher."
 
     ;; haskell
     (:type "module" :supports ("ag") :language "haskell"
-           :regex "^module\\s+JJJ\\b\\s+"
+           :regex "^module\\s+JJJ\\s+"
            :tests ("module Test (exportA, exportB) where"))
 
-                                        ; TODO Doesn't support any '=' in arguments. E.g. 'foo A{a = b,..} = bar'.
+    ; TODO Doesn't support any '=' in arguments. E.g. 'foo A{a = b,..} = bar'.
     (:type "top level function" :supports ("ag") :language "haskell"
-           :regex "^(JJJ\\b)(?!(\\s+::))\\s+((.|\\s)*?)=\\s+"
+           :regex "^JJJ(?!(\\s+::))\\s+((.|\\s)*?)=\\s+"
            :tests ("test n = n * 2"
                    "test X{..} (Y a b c) \n bcd \n =\n x * y"
                    "test ab cd e@Datatype {..} (Another thing, inTheRow) = \n undefined"
@@ -695,16 +695,16 @@ or most optimal searcher."
                  ))
 
     (:type "type-like" :supports ("ag") :language "haskell"
-           :regex "^\\s*((data(\\s+family)?)|(newtype)|(type(\\s+family)?))\\s+JJJ\\b"
+           :regex "^\\s*((data(\\s+family)?)|(newtype)|(type(\\s+family)?))\\s+JJJ\\s+"
            :tests ("newtype Test a = Something { b :: Kek }"
                    "data Test a b = Somecase a | Othercase b"
                    "type family Test (x :: *) (xs :: [*]) :: Nat where"
                    "data family Test"
                    "type Test = TestAlias"))
 
-    ; datatype contstuctor that doesn't match type definition
+    ; datatype contstuctor that doesn't match type definition.
     (:type "(data)type constructor 1" :supports ("ag") :language "haskell"
-           :regex "(data|newtype)\\s{1,3}(?!JJJ\\b)([^=]{1,30})=((\\s{0,3}JJJ\\b)|([^=]{0,500}?(\\|\\s{0,3}JJJ\\b)))"
+           :regex "(data|newtype)\\s{1,3}(?!JJJ\\b)([^=]{1,40})=((\\s{0,3}JJJ\\s+)|([^=]{0,500}?((?<!(-- ))\\|\\s{0,3}JJJ\\s+)))"
            :tests ("data Something a = Test { b :: Kek }"
                    "data Mem a = TrueMem { b :: Kek } | Test (Mem Int) deriving Mda"
                    "newtype SafeTest a = Test (Kek a) deriving (YonedaEmbedding)"
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -677,45 +677,69 @@ or most optimal searcher."
            :tests ("type test" "immutable test" "abstract test <:Testable" ))
 
     ;; haskell
-    (:type "function" :supports ("ag") :language "haskell"
-           :regex "^\\s*(let)?\\s*JJJ\\b\\s*(.+)?(?<==)"
-           :tests ("test n = n * 2" "let test x y = x * y")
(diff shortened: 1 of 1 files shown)


jacktasia:
Thanks for the PR! To answer your question: my understanding is `ag` does have multiline support. Can you expand on your tests that show otherwise? I think #102 was the last time this came up so @netromdk may know more than me.

Thanks again!

netromdk:
Yes, `ag` supports multiline matching. When I run your query on CLI then it works fine:
```
% ag "data.*{((.|\\s)*?::(.|\\s)*?,)*\\s*(test)\\b\\s*::(.|\\s)*}" test/test.hs
1:data Mem = Mem { mda :: A
2: , test :: Kek
3:  ,
4: aoeu :: E }
```
I put `data Mem = Mem { mda :: A\n , test :: Kek\n  ,\n aoeu :: E }` in the file.

It fails through the `:tests( .. )`, though:
```
TEST: ’data Mem = Mem { mda :: A
 , test :: Kek
  ,
 aoeu :: E }’
INPUT: ’data Mem = Mem { mda :: A
 , test :: Kek
  ,
 aoeu :: E }’
CMD: ’ag --nocolor --nogroup data.\*\{\(\(.\|\\s\)\*\?\:\:\(.\|\\s\)\*\?\,\)\*\\s\*\(test\)\\b\\s\*\:\:\(.\|\\s\)\*\}’
RESP: ’’
```

volhovm:
Alright, so as far as I've managed to figure out: ag doesn't support multiline when input is coming from stdin, so it has nothing to do with dumb-jump itself. IDK if it's possible to fix. Maybe we can change `ag` tests so they read things from the file directly (somehow, i'm not an elisp expert)?

netromdk:
I actually did the same thing for `git-grep`: `dumb-jump-run-git-grep-test`. Maybe it could just be renamed to something like `dumb-jump-run-test-from-file` and be used inside `dumb-jump-test-ag-rules`. But then the temp file should be named something more general, of course. But I haven't tested it for `ag` in this multiline situation.

netromdk:
Okay, if you do the following then it all works:

```
@@ -1212,9 +1212,9 @@ Optionally pass t for RUN-NOT-TESTS to see a list of all failed rules"
       (lambda (rule)
         (-mapcat
           (lambda (test)
-            (let* ((cmd (concat "ag --nocolor --nogroup "
+            (let* ((cmd (concat "ag --nocolor --nogroup --nonumber "
                                 (shell-quote-argument (dumb-jump-populate-regex (plist-get rule :regex) "test" 'ag))))
-                   (resp (dumb-jump-run-test test cmd)))
+                   (resp (dumb-jump-run-git-grep-test test cmd)))
               (when (or
                      (and (not run-not-tests) (not (s-contains? test resp)))
                      (and run-not-tests (> (length resp) 0)))
```

Except, of course, that `dumb-jump-run-git-grep-test` should be renamed as we talked about.

The deal was that while running this test it _cannot_ show the line numbers, like:
```
1:data Mem = Mem { mda :: A
2: , test :: Kek
3:  ,
4: aoeu :: E }
```
That's why it needs `--nonumber` only for running the tests.

volhovm:
Thank you, @netromdk. This is very helpful and I'll adopt it soon. I don't have a lot of time these days (was unsuccessfully testing my changes today) but I hope to finish this PR soon :)

volhovm:
So I figured out I had a configuration mistake: my `dumb-jump-force-searcher` was set to `auto` which defaulted to `git grep` (though it wasn't obvious at all, i thought `ag` is preferred since it's more powerful).
I've solved the tests problem (yay, thanks everybody) and I will work on refining regexps themselves a little bit more.

dumb-jump.el
```diff
@@ -1168,21 +1166,31 @@ If `nil` always show list of more than 1 match."
   "Use TEST as the standard input for the CMD."
   (with-temp-buffer
     (insert test)
-    (shell-command-on-region (point-min) (point-max) cmd nil t)
+    (shell-command-on-region (point-min) (point-max) cmd (current-buffer) t)
```
netromdk: Is it necessary to use `(current-buffer)`? I can see in the doc on `shell-command-on-region` that:
> If the value is nil, use the buffer ‘*Shell Command Output*’.

volhovm: Ok, right!

netromdk:
Looks great! :)

volhovm:
@netromdk is there any way i can match on something different from `JJJ`? Haskell types start with a capital letter, so `Command` rule should be different from `command`. I don't want some rules to fire up if substituted `JJJ` doesn't  follow the internal predicate itself. 

(i guess it's not possible, but asking anyway)

netromdk:
Well, "JJJ" is just substituted the search term but it's up to the regexp to worry about case sensitivity. `ag` uses smart case by default, for instance:
```
  -S --smart-case         Match case insensitively unless PATTERN contains
                          uppercase characters (Enabled by default)
```

I think most rules kind of assume case-insensitivity. But at least for `ag` it should match correctly given "Command" or "command" in your case, but that's only when the regexp contains either "Command" or "command" because otherwise it would just substitute in the search term which doesn't directly affect the rule around it's case-matching mode.

Off the top of my head; If we wanted to support this then the rule itself would have to be annotated with a case-matching mode, "like starts with capital letter", which is matched against the value substituting in for "JJJ". This annotation could actually be a regexp in itself to weed out which search terms are even tried with the rule.

Food for thought.

volhovm:
I'm a little bit stuck. I have a set of rules. Pattern `a` is matched by regex/rule `A`. I add another rule `B` (which is not related to `A`) and `a` is not matched by anything anymore. `b` is also not matched by `B` (but when the exact same string appears in tests, it is).

Any ideas what it can be? Maybe there's some magic about the way rules are matched?

netromdk:
That sounds weird, @volhovm. Do you mean it doesn't work with the inline rules but it works when you "jump" inside Emacs?

volhovm:
I think I'm actually done with this PR. Rules cover 99% of things i want, they work fast. I don't see any obvious problems.

dumb-jump.el
```diff
@@ -699,7 +699,7 @@ or most optimal searcher."
            :tests ("newtype Test a = Something { b :: Kek }"
                    "data Test a b = Somecase a | Othercase b"
                    "type family Test (x :: *) (xs :: [*]) :: Nat where"
-                   "data family Test"
+                   "data family Test "
```
netromdk: Is this space on purpose?

volhovm: Yes. I believe that every identifier should be followed by a space or eof. `\b` doesn't work because in haskell we have a lot of things like `Identifier` and `Identifier'`, so matching first will match second also.

jacktasia:
@volhovm Thanks for all the work on this. I'll be merging it as soon as I get back to a computer. @netromdk Thanks so much for reviewing and helping on this and everything else you've done for dumb jump

netromdk:
Oh, it's no problem. You're welcome :)

jacktasia:
Thanks again!

Issue #179, opened 2017-11-30
Should dumb-jump use (thing-at-point 'sexp) instead of (thing-at-point 'symbol) ?


ahungry:
Some very rough preliminary testing shows them to return the same results in a few different file types that I've checked, however in Common Lisp, this is a valid function call/name for a function titled `#/something`:

`(#/something "here")`

If the cursor is on the `#`, it won't find a valid symbol, while it will find the match if using the sexp type (I assume the sexp is just grabbing until the first whitespace.

Granted, this is quite unusual, especially considering CL uses the sharpsign as a way to reference functions like `#'symbol-name` as well.

Thoughts?

jacktasia:
Thanks for opening this! You're definitely on to something and this may be the proper solution for the likes of #168 and #174 (vs what I actually did). I _thought_ I had started with `sexp` then moved over to using `symbol` (for a reason I can't remember) but after a very quick look I can't find evidence of this.

I am definitely going to look closer into this tonight or this weekend and report back. Thanks again for opening this!

ahungry:
Thank you for a wonderful package and prompt response!

PR #181, opened 2017-12-04
thing-at-point 'sexp vs `symbol [WIP]

jacktasia:
Experiments related to #179 

I noticed that "things" `sexp` and `symbol` will behave differently with `thing-at-point` based on the current mode. Some example code below

```
(with-temp-buffer
  (funcall 'lisp-mode) ;; if removed or changed you can get different results for symbol vs sexp
  (let ((expected "#/something")) 
    (insert (concat "(" expected ")"))
    (goto-char (point-min))
    (forward-char 5)
    (let* ((cur (thing-at-point 'sexp)) ;; vs 'symbol
           (line (thing-at-point 'line))
           (matches (string= expected cur)))
           (message "thing: %s | matched: %s | line: %s" cur matches line))))
```

because of this I am considering adding two new options `dumb-jump-default-thing` and `dumb-jump-language-thing`. Where you can either define a "thing" per language or a default.

For anyone reading this please let me know if you have any feedback either way. Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ env:
   - EVM_EMACS=emacs-24.5-travis
   - EVM_EMACS=emacs-25.1-travis
   - EVM_EMACS=emacs-25.2-travis
+  - EVM_EMACS=emacs-25.3-travis
+
 script:
   - rg --version
   - ag --version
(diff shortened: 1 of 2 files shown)


netromdk:
I think that sounds reasonable. So you'll default both to `'symbol`?

jacktasia:
> I think that sounds reasonable. So you'll default both to 'symbol?

Yeah, basically I was thinking that `dumb-jump-default-thing` would default to `'symbol` and `dumb-jump-language-thing` would be an alist (I think?) and if the current language was not in there then it would fallback to the default.  Any thoughts/concerns? Thanks!

netromdk:
Sounds fine to me. Would be interesting with some meaningful tests showing the different cases probably.

Issue #186, opened 2017-12-14
Erroneously matches sub-words?


dleslie:
I noticed this behaviour just after submitting the Scheme support, but it matches the existing behaviour found in the elisp support.

I think it's best described in code:
https://github.com/dleslie/dumb-jump/commit/47bd47654e5049c83a5cc2aa35747e34b81f85c0

Should those new `not` tests pass? They do not at present, and I suspect it's because `\\j` word boundaries aren't working as intended. Unless the intent is to match sub-words. ;)

jacktasia:
Thanks for opening this! This is probably related to this: #181 and #179 That is, it's about what emacs is getting at point. I think. You can verify by turning on the debug flag which is described in the readme. Please let me know what you find. Thanks!

dleslie:
I was under the impression the `not` tests were applied directly to the underlying search provide, be it `rg`, `ag` or other? Anyhow...

For this:
`(defun tester () ...)`

`thing-at-point` will be `tester` for either `sexp` or `symbol`; but I'm looking for the definition of `test`. I can see people wanting this sort of result to show up, however, because they may be trying to recall the full name of a function that they can only recall the prefix for. Hence the bug in the form of a question. 😄

jacktasia:
Ohhh, I see now. Sorry, when I first read this it was on my phone and clearly read it too fast and assumed it was the issue(s) I mentioned. I see what you're saying now and will look closer soon.

netromdk:
This issue can be closed now, right @jacktasia?

jacktasia:
@netromdk Yeah, that's my understanding. I am closing it. @dleslie Please let me know if you're still running into issues.

PR #188, opened 2017-12-15
Fix bad boundary matches

jacktasia:
#186

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -213,11 +213,12 @@ or most optimal searcher."
   :type 'boolean)
 
 (defcustom dumb-jump-find-rules
-  '((:type "function" :supports ("ag" "grep" "rg" "git-grep") :language "elisp" :regex "\\\((defun|cl-defun)\\s+JJJ\\j"
+  '((:type "function" :supports ("ag" "grep" "rg" "git-grep") :language "elisp"
+           :regex "\\\((defun|cl-defun)\\s+JJJ\\j"
            ;; \\j usage see `dumb-jump-ag-word-boundary`
            :tests ("(defun test (blah)" "(defun test\n" "(cl-defun test (blah)" "(cl-defun test\n")
-           :not ("(defun test-asdf (blah)" "(defun test-blah\n" "(cl-defun test-asdf (blah)" "(cl-defun test-blah\n"))
-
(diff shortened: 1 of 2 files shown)


dumb-jump.el
```diff
@@ -221,13 +221,19 @@ or most optimal searcher."
...
     (:type "variable" :supports ("ag" "grep" "rg" "git-grep") :language "elisp"
-           :regex "\\\(defcustom\\b\\s*JJJ\\j" :tests ("(defcustom test " "(defcustom test\n"))
+           :regex "\\\(defcustom\\b\\s*JJJ\\j"
+           :tests ("(defcustom test " "(defcustom test\n")
+           :not ("(defcustom tester"))
+
 
     (:type "variable" :supports ("ag" "grep" "rg" "git-grep") :language "elisp"
-           :regex "\\\(setq\\b\\s*JJJ\\j" :tests ("(setq test 123)") :not ("setq test-blah 123)"))
+           :regex "\\\(setq\\b\\s*JJJ\\j" :tests ("(setq test 123)")
```
netromdk: Like in the other places, it would be nice with a newline before `:tests` for readability.

dumb-jump.el
```diff
@@ -125,25 +125,25 @@ or most optimal searcher."
   :type 'string)
 
 (defcustom dumb-jump-ag-word-boundary
-  "(?![\\w\\?\\*-])"
+  "(?![a-zA-Z0-9\\?\\*-])"
```
netromdk: The motivation for using `a-zA-Z0-9` instead of `\w` is that you don't want it to match for `_` which is the only other character not matched by this change?

jacktasia: For _some_ searchers it was causing bad matches as outlined in #186 . For example, saying `tester` was a match when jumping on `test`. You can see evidence of this in the early commits for this PR with the :x:s, but here's a deep link to one example:  https://travis-ci.org/jacktasia/dumb-jump/jobs/316563561 
Thanks again for looking at this!

jacktasia: This leads me to believe for these searchers with the options we give them `\w` is not recognized at all.

netromdk: Very weird but I see the point. Np :)

jacktasia:
@netromdk Yup, thanks!

PR #191, opened 2017-12-30
Fixes #137 Support tramp

### LINKED ISSUE #137: Support tramp
Works on a emacs running on remote, but doesn't work over tramp. 

Host: GNU Emacs 25.1.1 (x86_64-apple-darwin13.4.0, NS appkit-1265.21 Version 10.9.5 (Build 13F1911)) of 2016-09-17

Remote: GNU Emacs 24.5.1 (x86_64-pc-linux-gnu) of 2016-04-17 on lgw01-04, modified by Debian

The only difference I can see between working and not working is tramp.

popoppo:
As for #137,
when a remote file is accessed with tramp, `dumb-jump-get-project-root` returns a remote path such as "/ssh:..." and this is used as a project root in `dumb-jump-run-command`, but search commands need a local path as a target, so the search will fail with the error "No such file or directory".

On the other hand, `find-file` (in `dumb-jumb-goto-file-line`) needs a remote path to jump to the target file.

This PR changes some parts of code to handle the project root properly, and also adds tests for them.
However, as @jacktasia said, there would be many ways to use tramp, so I'm not sure if that is enough.

Anyway, could you look over the changes? > @jacktasia 
Thanks. 🎄

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -1990,8 +1990,7 @@ searcher symbol."
   "Run the grep command based on the needle LOOK-FOR in the directory TOSEARCH"
   (let* ((proj-root (if (file-remote-p proj)
                         (directory-file-name
-                         (tramp-file-name-localname
-                          (tramp-dissect-file-name proj)))
+                         (tramp-file-name-localname (tramp-dissect-file-name proj)))
                       proj))
          (cmd (funcall generate-fn look-for cur-file proj-root regexes lang exclude-args))
          (shell-command-switch (dumb-jump-shell-command-switch))
@@ -2000,7 +1999,6 @@ searcher symbol."
     (when dumb-jump-debug
       (dumb-jump-message
        "-----\nDUMB JUMP DEBUG `dumb-jump-run-command` START\n----- \n\ncmd: \n\t%s\n\nraw results: \n\n\t%s \n\n-----\nDUMB JUMP DEBUG `dumb-jump-run-command` END\n-----\n" cmd rawresults))
-
     (when (and (s-blank? rawresults) dumb-jump-fallback-search)
       (setq regexes (list dumb-jump-fallback-regex))
       (setq cmd (funcall generate-fn look-for cur-file proj-root regexes lang exclude-args))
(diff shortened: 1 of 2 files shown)


popoppo:
NOTE:
This feature won't work in a situation such as

- a.txt on localhost (has ag)
- b.txt on a remote server (doesn't have ag)

That's because,,,
when you use `dumb-jumb-go` for a.txt, `dumb-jump--ag-installed?` will have `t` and that's also used when you call `dumb-jumb-go` for b.txt, but the remote server doesn't have ag and it ends in failure.

jacktasia:
@popoppo Thanks for the PR! This is great. Please give me some more time to look this over more closely. before I merge. 

@netromdk Do you have any concerns with this?

jacktasia:
Sorry for the delay on this. When I was testing this locally the tests were failing for me and then when I had travis re-run the tests they also started to fail...which seems to be due to a change with a dependency: https://github.com/rejeep/f.el/commit/55ae23e4a1e26d0c74568da2d8997196afef7d32 which kind of confuses me since I am pinning the version.

Tonight I'll update the tests and then complete the review of the PR.

netromdk:
Btw [it seems](https://travis-ci.org/jacktasia/dumb-jump/jobs/323166212#L1066) trailing slashes are now being added from somewhere?

Examples:

```
Test dumb-jump-find-proj-root-test condition:
    (ert-test-failed
     ((should
       (string= found-project test-data-dir-proj1))
      :form
      (string= "/home/travis/build/jacktasia/dumb-jump/test/data/proj1/" "/home/travis/build/jacktasia/dumb-jump/test/data/proj1")
      :value nil))

Test dumb-jump-quick-look-test condition:
    (mock-error
     (popup-tip "/src/js/fake.js:3: function doSomeStuff() {")
     (popup-tip "src/js/fake.js:3: function doSomeStuff() {"))
```

jacktasia:
@netromdk Thanks for looking this over!

> Btw it seems trailing slashes are now being added from somewhere?

This isn't related to this PR. The tests for it originally all passed, but I re-triggered them on TravisCi when they didn't pass locally. See my previous comment on this thread (seems to be a breaking change from f.el). If you remove your local `.cask` in `dumb-jump` locally and re-run `cask` and `make test` on master the same thing will happen. If it doesn't please let me know.

jacktasia:
Sorry for the delay in getting this merged. I want to get #193 merged in first since it will make these tests pass again. The f.el dependency had some backwards incompatible changes that just happened to have been released right around the time of this PR.

popoppo:
No problem. I'm not in a hurry at all 🍵

netromdk:
Seems good to go now, @jacktasia?

jacktasia:
> Seems good to go now, @jacktasia?

I just want to do a couple of manual tests and see how easy it would be for me to increase the coverage, then I'll merge (probably within the next 24 hours).

popoppo:
Thank you guys!!

PR #193, opened 2018-01-07
handle recent changes to f.el

jacktasia:


jacktasia:
Thanks for the review @netromdk!

Miciah:
This change seems strange.  Does it fix some problem?

I think it causes at least one problem.  `dumb-jump--format-result` removes the project root from results, but because the project root doesn't have a trailing slash, the formatted results end up with a leading slash.

I noticed that `dumb-jump-generate-ag-command` and `dumb-jump-generate-rg-command` both use `file-name-as-directory` to add the slash back to the project root.  Perhaps `dumb-jump--format-result` should do the same (and I'd be happy to make a PR if you think it should), but wouldn't it make more sense for `dumb-jump-get-project-root` to return a directory name?

jacktasia:
> This change seems strange. Does it fix some problem?

Yes, `f.el` made a backwards incompatible change which made it necessary since you can't define a max version for dependencies (and most people use MELPA unstable) it was causing issues for anyone on a fresh install because they would get the backwards incompatible change and it would break dumb-jump.

All this PR does is require the new version and make it so it behaves the same way as it did before.

> I think it causes at least one problem.

Any problems were probably already present. I am more than happy to accept any PRs that will improve things.

>  but wouldn't it make more sense for dumb-jump-get-project-root to return a directory name?

I am not sure I understand this question. As in not an absolute path?

Miciah:
> All this PR does is require the new version and make it so it behaves the same way as it did before.

It also adds `(s-chop-suffix "/" ...)`.  Was that accidental?

> > but wouldn't it make more sense for dumb-jump-get-project-root to return a directory name?

> I am not sure I understand this question. As in not an absolute path?

I mean with the trailing slash that the `s-chop-suffix` removes.

jacktasia:
> It also adds (s-chop-suffix "/" ...). Was that accidental?

No, `f-expand` behaved this way before and then they made a change where it did not...and since other code operated under the assumption that it wouldn't be there this was the fix...

Miciah:
I see.  So I am complaining in the wrong place (sorry to victimize an innocent PR!); but it still seems odd to remove the trailing slash here and then add it back in places that treat it as a directory name.

jacktasia:
> I see. So I am complaining in the wrong place (sorry to victimize an innocent PR!); 

Haha, no problem! I appreciate the interest.

> but it still seems odd to remove the trailing slash in dumb-jump-get-project-root and then add it back in places that treat it as a directory name.

I don't necessary disagree. This was an urgent fix at the time to just get the same functionality it had before. Like I said before, if there's any issues then they were probably always there.

PR #199, opened 2018-01-22
fix line parsing error in dumb-jump-helm-persist-action

lewang:
Fix problem in parsing line removing the line-number portion during `s-split`.

```emacs-list
(let* ((match "vault.clj:222: (defn status")
       (parts (--remove (string= it "")
                        (s-split "\\(?:^\\|:\\([0-9]+\\):\\)"  match))))
  parts)
;;; => ("vault.clj" " (defn status")

```

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -1395,11 +1395,9 @@ Optionally pass t for RUN-NOT-TESTS to see a list of all failed rules"
 
 (defun dumb-jump-helm-persist-action (match)
   "Previews a MATCH in a temporary buffer at the matched line number when pressing \\<keymap>C-j</keymap> in helm."
-  (let* ((parts (--remove (string= it "")
-                          (s-split "\\(?:^\\|:\\)[0-9]+:"  match)))
-         (file-line-part (s-split ":" (nth 0 parts)))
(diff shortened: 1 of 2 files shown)


netromdk:
Thanks for the PR, @lewang. But coverage is down 16% and lots of tests are not passing. Please run the tests locally to fix the issues.

netromdk:
Great, @jacktasia  👍

jacktasia:
@lewang Thanks for the PR! It looks like the test just needed to be updated, which I went ahead and pushed into this PR via aaf9562.

@netromdk Thanks, it looks like the large drop in coverage was due to the tests timing out and thus only got ~80% through the tests but Travis still dutifully send stats to coveralls anyway.

Assuming the tests pass this seems good to me now, but I don't actually use helm or this feature so if one or both of you can confirm this still works as expected then I am cool with it being merged. Thanks again guys!

netromdk:
I will test it out.

netromdk:
The persist action fails:
```el
let*: Wrong type argument: stringp, nil
```
Trying to track it down.

jacktasia:
Thanks @lewang and @netromdk!

Issue #25, opened 2016-05-01
Save before jump


adouzzy:
Just wondering why is this necessary. It's not very convenient.

jacktasia:
It's necessary because `ag`/`grep` search saved files. If you've changed a file a lot and try to jump then you could very well jump to an old position, which is extremely confusing and makes dumb-jump seem broken.

nickdrozd:
@jacktasia I spent several minutes yesterday thinking `dumb-jump` was busted because it kept dropping me off four or five lines above the functions I was searching for. Then I realized the file wasn't saved, and I thought "That's stupid, why doesn't it prompt to save before searching?" Then I came here and discovered that someone else had the exact opposite complaint ("This sucks, why does it keep prompting me to save?").

You can't please everyone I guess.

Or can you? Maybe there could be a prompt-before-searching flag? Is there one already? I don't know anything about the history of this issue. Was the prompt dropped at some point? If so, could you point me to the commit? I would be happy to add a customizable option if you think it would useful.

It's a great package, btw. Aside from this very minor problem, it really does "just work".

jacktasia:
@nickdrozd Are you on an old version? [It should be warning you when you jump to an unsaved buffer](https://github.com/jacktasia/dumb-jump/blob/master/dumb-jump.el#L1883), but maybe there's a bug in some case. I know I see this prompt pretty often, but if you know of a case where this failing I'd love to hear about it. For the record, here's the PR/commit where this changed: https://github.com/jacktasia/dumb-jump/pull/141

> It's a great package, btw. Aside from this very minor problem, it really does "just work".

Thanks! I appreciate the kind words.

nickdrozd:
@jacktasia This problem is on master. I took a look and I think I understand what is happening.

The function that checks for modifications is 

```
(defun dumb-jump-file-modified-p (path)
  "Check if PATH is currently open in Emacs and has a modified buffer."
  (let ((modified-file-buffers
         (--filter
          (and (buffer-modified-p it)
               (buffer-file-name it)
               (file-exists-p (buffer-file-name it)))
          (buffer-list))))
    (member path (--map (buffer-file-name it) modified-file-buffers))))
```

Now, when I test this code manually, it works fine as long as I pass
in full file paths. But it doesn't get full paths, just the filename
(`dumb-jump.el`, rather than `/path/to/dumb-jump.el`).

Why is the short filename getting passed in? The actual command that
does the searching (`git grep` for me) is executed in
[...]

PR #200, opened 2018-02-15
fix jumping to unsaved buffer bug when using git-grep

jacktasia:
#25


nickdrozd:
That does it, thanks!

PR #218, opened 2018-05-05
Add basic support for Pascal

murmour:
Self-descriptive.


dertuxmalwieder:
Is there a reason why Pascal is not in the list of supported languages just yet though? :-)

dertuxmalwieder:
Good - thank you too!

Issue #27, opened 2016-05-26
Please add support to PL/SQL procedures and packages


zsoltika:
In the same (projectile) directory tree.

Would be great to jump to `package_name.procedure_name` or `package_name.function_name`, and `package_name.constant_name`, where `package_name` is usually a `package_name.(prc|pks|sql)` file in the actual project directory tree.

PR #221, opened 2018-05-27
Add basic support for jumping to functions in SQL files

deuill:
This is incredibly basic, and adds support for syntax generally defined by the SQL standard, with
provisions for syntax exposed by popular database engines, such as MySQL, PostgreSQL, MS-SQL etc.

Both upper-case and lower-case definitions will match. A better solution would'be been to allow for
case-insensitive matching, but mixed case within the same word should be rare enough to not be a
problem at this stage.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -1044,6 +1044,25 @@ or most optimal searcher."
                    "CREATE PROCEDURE test (OUT p INT)"
                    "create definer = 'test'@'localhost' procedure test()"))
 
+    (:type "table" :supports ("ag" "grep" "rg" "git-grep") :language "sql"
+           :regex "(CREATE|create)\\s+(.+?\\s+)?(TABLE|table)(\\s+(IF NOT EXISTS|if not exists))?\\s+JJJ\\b"
+           :tests ("CREATE TABLE test ("
+                   "create temporary table if not exists test"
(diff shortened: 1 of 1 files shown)


deuill:
Closes #27

deuill:
I tried my hand at implementing support for variables, but the syntax is somewhat different between database engines (at least, the ones I'm familiar with).

I'll look at implementing support for jumping to tables, views, and types, however.

dumb-jump.el
```diff
@@ -1036,6 +1036,14 @@ or most optimal searcher."
                    "functor test (T:TEST) ="
                    "functor test(T:TEST) ="))
 
+    ;; sql
+    (:type "function" :supports ("ag" "grep" "rg" "git-grep") :language "sql"
+           :regex "(CREATE|create)\\s+(.+?\\s+)?(FUNCTION|function|PROCEDURE|procedure)\\s+JJJ\\s*\\\("
```
deuill: I hate having to define the two variants, but couldn't find support for declaring the regex as case-insensitive. Perhaps adding an additional key for additional options to be passed to the command would do?

deuill:
Review away!

jacktasia:
This is awesome. Thanks so much for the PR!

deuill:
Thanks for building this awesome package ;)

PR #222, opened 2018-05-31
Simplify helm usage and fix persistent action

Miciah:
Before this commit, `dumb-jump-prompt-user-for-choice` would provide helm only with the formatted choices.  The persistent action would need to parse the selected choice, which did not always work properly because the path may be truncated in the formatted choice.

* `dumb-jump.el` (`dumb-jump-prompt-user-for-choice`): Instead of invoking `helm` and then invoking `dumb-jump-to-selected` on the result, invoke `helm` with a default action that invokes `dumb-jump-result-follow`.
Pass in the actual results along with the formatted choices so that actions have actual results to work with.

* `dumb-jump.el` (`dumb-jump-helm-persist-action`): Expect the actual result rather than the formatted choice.
Fix the doc string.

* `test/dumb-jump-test.el`
(`dumb-jump-prompt-user-for-choice-correct-helm-persistent-action-test`):
Adjust test input per the changes to `dumb-jump-helm-persist-action`.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -1512,11 +1512,12 @@ Optionally pass t for RUN-NOT-TESTS to see a list of all failed rules"
         (when result
           (dumb-jump-result-follow result))))
 
-(defun dumb-jump-helm-persist-action (match)
-  "Previews a MATCH in a temporary buffer at the matched line number when pressing \\<keymap>C-j</keymap> in helm."
-  (let* ((line-parts (dumb-jump-parse-response-line match "."))
-         (file (nth 0 line-parts))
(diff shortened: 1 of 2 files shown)


jacktasia:
Thanks for the PR! Seems like there's an [issue with a test](https://travis-ci.org/jacktasia/dumb-jump/jobs/386444635).

jacktasia:
> Sorry, ERT is kind of new to me. Tests are now updated. 

No worries, thanks!

> I wonder whether there's a more comprehensive way to test the helm code.

Yeah, probably. I am happy to accept PRs if you have any ideas. I always do a fair amount of manual testing when things like this change.

PR #227, opened 2018-08-18
Add PHPDoc's @method and @property tag for PHP magic method

zonuexe:
These notations are defined by phpDocumentor.

- http://docs.phpdoc.org/references/phpdoc/tags/method.html
- http://docs.phpdoc.org/references/phpdoc/tags/property.html


jacktasia:
Thanks!

PR #231, opened 2018-09-28
Find Fortran subroutines defined in interface blocks

billsacks:
This finds generic procedures defined like this:

```Fortran
  interface endrun
     module procedure endrun_vanilla
     module procedure endrun_globalindex
  end interface
```

(where you're looking for the definition of 'endrun').

Since there is no instance of `subroutine endrun (...)` (only `subroutine endrun_vanilla (...)` and `subroutine endrun_globalindex (...)`), this change is needed in order to find the place where `endrun` is defined.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -671,6 +671,11 @@ or most optimal searcher."
            :tests ("function test (foo)" "integer function test(foo)" "subroutine test (foo, bar)")
            :not ("end function test" "end subroutine test"))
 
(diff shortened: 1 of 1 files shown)


billsacks:
Thanks a lot for this very useful package!

Some caveats:

1. I wasn't able to get cask to work, so wasn't able to run the tests on this change

2. I have only been using this package for a day, and this change for a few minutes, so it's possible that this change does the wrong thing in some situations. If you prefer to wait to merge this until I have more experience with the change, that's fine.

PR #235, opened 2018-10-19
Added Fsharp support

walseb:


jacktasia:
Thanks for the PR! It looks like the `rg` tests are failing. If you don't want to fix them for `rg` it's totally fine to just drop `rg` from the `:supports` block. Thanks again!

walseb:
That's really weird because `rg --type-list` returns `fsharp: *.fs, *.fsi, *.fsx`.
Anyway, I'm removing support for it like you suggested.

walseb:
Oh yeah, the problem could also be that rg has some special regex rule. Anyway rg support isn't that important to me.

walseb:
Oh, I forgot to remove some other rg support flags, fixing

PR #236, opened 2018-10-21
Call ivy-read canonically

akirak:
I'd better suggest you pass an action function to `ivy-read` as `:action` argument rather than pass the result of `ivy-read` to the action. This allows navigating through candidates by `C-M-m` without closing the prompt. 

Setting `:caller` is also recommended, as it allows per-function customization of Ivy.


akirak:
Okay, I'll work on it later.

jacktasia:
Thanks for the PR!

jacktasia:
Closed in favor of #308

PR #238, opened 2018-10-22
Jump to Java/C#/Vala functions that return array types

Wilfred:
Previously we excluded [] in our regex, so code of the form:

   public int[] foo()

was not found.


jacktasia:
Thanks for the PR! Just waiting for github webhooks to come back fully so CI can run the tests: https://status.github.com/messages

PR #244, opened 2018-11-29
support ruby namespaced classes, closes #243

### LINKED ISSUE #243: Jumping to namespaced classes doesn't work in Ruby
In Ruby, the conventional way to namespace a class is to nest it in one or more modules:

```ruby
module Health
  module Checks
    class QueueGrowth
      [code for checking queue growth]
    end
  end
end
```

(For the record, there's a less conventional, worse way of writing this code that I don't think Dumb Jump needs to support:

```ruby
class Health::Checks::QueueGrowth
end
```
)

You can then reference the class like this:

* anywhere: `::Health::Checks::QueueGrowth`
* anywhere without a conflicting namespace: `Health::Checks::QueueGrowth`
* inside the `Health` module: `Checks::QueueGrowth`
* inside the `Health::Checks` module: `QueueGrowth`

`dumb-jump-go` usually doesn't know what to do when you call it with any of those references I just listed. It either jumps to the very thing your cursor is over (if it's the only reference of its type), or it presents a list of other references to the class, but not the class itself:

image

It would be a bit of work to make Dumb Jump know about the intricacies of Ruby namespacing, but I think things would be greatly improved simply by looking only for the bit after all the colons (`QueueGrowth` in the example above), and presenting a list of all the places that `module QueueGrowth` or `class QueueGrowth` appear in the code.

The code base I work in uses namespaces extensively, and I usually end up copying the bit after the colons, calling `dumb-jump-go-prompt`, and pasting.

jacktasia:
#243

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -1746,6 +1746,8 @@ to keep looking for another root."
     (nth 1 (s-split "/" look-for)))
    ((and (or (string= lang "ruby") (string= lang "crystal")) (s-starts-with? ":" look-for))
     (s-chop-prefix ":" look-for))
+   ((and (string= lang "ruby") (s-contains? "::" look-for))
+    (-last-item (s-split "::" look-for)))
    ((and (string= lang "systemverilog") (s-starts-with? "`" look-for))
     (s-chop-prefix "`" look-for))
    (t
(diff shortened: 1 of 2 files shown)


jacktasia:
@patbl This look correct to you?

patbl:
Yes, that looks good. Thanks so much!

PR #246, opened 2018-11-29
Perl: Support the new syntax of function declarations

Vifon:
I've added the support for the new style function declarations with the argument list just after the function name. I also removed a stray backslash in the regex that was there: `\\\{` doesn't make sense in an Elisp string.


Vifon:
It should also work if somebody uses [the prototypes](https://perldoc.perl.org/perlsub.html#Prototypes). This syntax shouldn't clash with function calls as the call never contains the token `sub`.

jacktasia:
Awesome, thanks!

PR #248, opened 2018-12-21
Ruby: support for fully qualified constants

dustym:
Adds support for fully qualified constants like "::Foo". Currently dumb-jump peeks at the first character of the symbol to see if it starts with a colon and thus is a ruby symbol, then strips the colon. This patch checks to see if the symbol starts with two colons and, if so, doesn't strip the colon.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -1751,7 +1751,7 @@ to keep looking for another root."
   (cond
    ((and (string= lang "clojure") (s-contains? "/" look-for))
     (nth 1 (s-split "/" look-for)))
-   ((and (or (string= lang "ruby") (string= lang "crystal")) (s-starts-with? ":" look-for))
+   ((and (or (string= lang "ruby") (string= lang "crystal")) (and (s-starts-with? ":" look-for) (and (not (s-starts-with? "::" look-for)))))
     (s-chop-prefix ":" look-for))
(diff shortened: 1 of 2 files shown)


dumb-jump.el
```diff
@@ -1751,7 +1751,7 @@ to keep looking for another root."
   (cond
    ((and (string= lang "clojure") (s-contains? "/" look-for))
     (nth 1 (s-split "/" look-for)))
-   ((and (or (string= lang "ruby") (string= lang "crystal")) (s-starts-with? ":" look-for))
+   ((and (or (string= lang "ruby") (string= lang "crystal")) (and (s-starts-with? ":" look-for) (and (not (s-starts-with? "::" look-for)))))
```
jacktasia: The second `and` here is unnecessary. That said, now that I understand the issue you're fixing I am pretty sure we don't need to actually add anything and we can simply swap the conditional blocks so it looks like this:

```
   ((and (string= lang "ruby") (s-contains? "::" look-for))
    (-last-item (s-split "::" look-for)))
   ((and (or (string= lang "ruby") (string= lang "crystal")) (s-starts-with? ":" look-for))
    (s-chop-prefix ":" look-for))
```

Does that seem right to you?

jacktasia:
Thanks for the PR and report! I made one small comment.

jacktasia:
Thanks!

PR #261, opened 2019-02-17
Add nix support

terlar:
This is very basic, but nix is also a very basic language. Everything is basically variables (definitions) and they could be either data or functions.

The recommended way to find definitions got me to think about this:
https://nixos.wiki/wiki/Nix_Expression_Language#Finding_a_.28function.2Fpackage.2F.E2.80.A6.29_definition

I haven't tested this much yet, but seems to work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ There is currently basic support for the following languages:
 * LaTeX
 * Lua
 * Nim
+* Nix
 * Objective-C
 * OCaml
 * Pascal
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -523,6 +523,12 @@ or most optimal searcher."
            :tests ("type test = object" "type test {.pure.} = enum")
            :not ("type testnot = object"))
 
+    ;; nix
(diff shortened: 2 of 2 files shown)


jacktasia:
Thanks for the PR! It looks like the tests are failing for `grep` and `git-grep`. I can dig in more after work today.

terlar:
Yes, I saw that too, haven't had time to look at it yet either. Sorry about that. Yes please, if you have time. Otherwise I will look at it later today or tomorrow :)

jacktasia:
Ah, it's because you have a `\n` in the test and `grep` and `git-grep` do not support multi-line matching. I pushed a fix into your branch. Please let me know what you think.

terlar:
Ah, that explains it. I think that is fine, I personally don't use that, but perhaps it would be neat to have support for them as well. I am not sure if the match needs to be like that or if the match should just be on the definition line ending with equals might be okay. As long as it would work for definitions with the value on the next line

terlar:
Thank you! For the assistance and a great package. Will try to get cask installed until next time so I could run the tests and make a more complete PR. :)

PR #265, opened 2019-03-05
Fix git grep with default recurse submodules

larkery:
`git grep`'s `--untracked` argument is incompatible with `--recurse-submodules` and the `.gitconfig` option `submodule.recurse = true` makes `--recurse-submodules` the default.

As a result if you have `submodule.recurse = true` in your `.gitconfig`, dumb jump will silently fail to work in any git repository that has submodules checked out, unless you disable `dumb-jump-git-grep-search-untracked`. This change makes `dumb-jump-git-grep-search-untracked` imply `--no-recurse-submodules`, which will defeat the config option and make it work, albeit ignoring the content of the submodules.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -2522,7 +2522,7 @@ Using ag to search only the files found via git-grep literal symbol search."
          (cmd (concat dumb-jump-git-grep-cmd
                       " --color=never --line-number"
                       (if dumb-jump-git-grep-search-untracked
-                          " --untracked"
+                          " --untracked --no-recurse-submodules"
                         "")
                       " -E"))
          (fileexps (s-join " " (--map (shell-quote-argument (format "%s/*.%s" proj it)) ggtypes)))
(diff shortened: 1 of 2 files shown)


jacktasia:
Thanks for the PR! I can't see TravisCI from where I am right now, but I am guessing the failing test is related to this line in the tests: https://github.com/jacktasia/dumb-jump/blob/master/test/dumb-jump-test.el#L159 That is, you'll have to add `--recurse-submodules` there too after `--untracked` too.

larkery:
I'll make that change - I confess I just hacked the relevant line and ignored the tests in a fit of extreme professionalism.

larkery:
I've fiddled with it, but it looks like travis is broken in some other way, checksum is bad for a package list or something.

jacktasia:
Closing this in favor of #268 just in case this introduces unknown behavior for other people or other args need to get added in the future.

Issue #250, opened 2019-01-25
Question: Multi-line patterns with git-grep+ag?


mhcerri:
Hi,

Do you think it would possible to implement a way to first search simply by the symbol name using `git grep` and then use the set of files returned to do an `ag` search with the full pattern?

jacktasia:
Hi @mhcerri 

That's certainly possible, but I don't think it would be a trivial change. Can you provide some details on why you want/need that functionality?

mhcerri:
I would love to use it on the linux kernel source tree. Multi-line declarations are very common there and anything besides `git grep` is very slow.

I can try to put together a pull request for that. Do you have any suggestions or directions on how to proceed?

jacktasia:
> Multi-line declarations are very common there and anything besides git grep is very slow.

Gotcha, thanks for the context.

> Do you have any suggestions or directions on how to proceed?

I think so. Let me think about this for a bit and get back to you. It's possible there's a couple of places to hook into that would make this really easy.

jacktasia:
@mhcerri I apologize for the delay on getting back to you on this. I think I have come up with a pretty good way to hook in the functionality you have suggested. That said, I would like to test it against something real before I concert my bash test into Emacs Lisp. Can you please recommend a good function with a multi-line declaration in the kernel to test against? By "good" I mean whatever criteria that makes sense to you but one that you would think would be particularly slow for whatever reason would probably be best. Thanks!

jacktasia:
Thanks for the info! I am going to try to implement and PR this soon.

jacktasia:
@mhcerri I got something seemingly working well in #254  Can you please try it for your needs and give me any feedback you may have there? Thanks!

mhcerri:
Hi, @jacktasia.

I tested several symbols and it works perfectly so far. It was able to find multi-line signatures way faster. I also tested it on smaller repos and it works flawless as well, returning immediately.

Thank you a lot!

jacktasia:
@mhcerri Awesome! Thanks for trying it out. I am going to keep the PR open until I can add some more tests then I'll merge. Please let me know if you run into anything. Thanks!

PR #271, opened 2019-03-07
Added search argument options for each search tool

DeeeeLAN:
Hi, I took a stab at adding the search arguments myself. I made the git-grep one from yesterday more generic, and added options for ag and rg.


DeeeeLAN:
Sorry this took a little churn. I don't have cask locally.

jacktasia:
Thanks for opening this! I think `dumb-jump-git-grep-search-untracked` needs to stay as a boolean option for backwards compatitlity with people's configuration since it has existed for a while. Although I agree that `dumb-jump-git-grep-search-args` can replace `dumb-jump-git-grep-search-untracked-args` to just become a totally new feature. Thoughts?

DeeeeLAN:
Yeah I guess we don't want to break anybody's code. I can update it tomorrow or you can take care of it now, whatever works.

jacktasia:
Okay, cool. I pushed up a commit with the changes I think make sense. I'll probably merge this sometime tomorrow unless you see anything else you think should change. Thanks again!

DeeeeLAN:
That looks good to me, I just fixed a couple comments you missed. I noticed that there is a git-grep+ag function- I don't know if it makes sense to add any flags from either or both tools to that function or not.

jacktasia:
> I just fixed a couple comments you missed.

Great, thanks!

> I noticed that there is a git-grep+ag function- I don't know if it makes sense to add any flags from either or both tools to that function or not.

Yeah, this is a specialized/experimental searcher for developing on the linux kernel (#250 / #254) that I don't think needs to be updated for this (yet).

PR #287, opened 2019-08-30
Add option to jump to modified buffers without confirmation

codesections:
Thanks for the dumb-jump—it's a great package and the simplicity really appeals to me.  This is a PR for a patch that I find personally useful and thought others might as well.

It adds the option to jump directly to a buffer without being prompted for confirmation, even if that buffer is modified.  I find that this is more often what I want, especially when jumping back up to the definition of a function defined earlier in the file (which, if I have been editing lower in the file, won't have moved at all).

Specifically, if the `dumb-jump-confirm-jump-to-modified-file` variable is nil, then jumping to a modified buffer will print a warning message but will not wait for confirmation.

I also revised the README to explain this additional option.


jacktasia:
Thanks for opening this! I'll take a look after work today.

PR #292, opened 2019-09-16
Add support for OpenSCAD scad files.

Lenbok:
OpenSCAD is a 3D modelling tool at: https://www.openscad.org/

In emacs, we can use http://melpa.org/#/scad-mode to assist editing, but it's nice to have dumb-jump providing definition jumping, so here we go.


jacktasia:
Thanks for opening this! I'll take a closer look tonight after work.

PR #300, opened 2019-09-27
Restrict ag searching based on extension for languages without an agtype

Lenbok:
In my case with OpenSCAD, `ag` does not have built in support. The `ag` command that `dumb-jump` builds ends up searching through all files in the project (which includes large 3D model files), resulting in very poor performance.

This patch tells `ag` to use the file extensions configured for the language if there is no agtype. I guess a similar thing could be done for other searching back-ends.


jacktasia:
Thanks for opening this! I'll take a closer look after work tonight.

Issue #211, opened 2018-03-27
ivy selector doesn't support ivy-call


amosbird:
Executing `ivy-call` on candidate has no effect at all.

jacktasia:
Thanks for opening this.

> Executing ivy-call on candidate has no effect at all.

Can you please expand on what you're expecting this to do? I assume keep the candidate list up but still jump to it? Thanks!

amosbird:
yup, like previewing the candidate. `counsel-rg` supports that.

jacktasia:
I think #236 will address this.

PR #308, opened 2019-11-09
Assign an action for dumb-jump-ivy-jump-to-selected.

leungbk:
This allows for ivy-resume to work with dumb-jump.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -1864,11 +1864,13 @@ This is the persistent action (\\[helm-execute-persistent-action]) for helm."
           (plist-get result :line)
           (s-trim (plist-get result :context))))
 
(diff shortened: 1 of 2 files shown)


leungbk:
Oh, I didn't notice #236, which should also fix the missing integration for `ivy-resume`. I assume that PR failed CI for the same reason. I'll have to learn how `el-mock` works within the next few days.

leungbk:
OK, I'm not clear on all the details of `el-mock`, but i think it's OK to remove the `(mock (dumb-jump-result-follow` since it's now taking place in the `:action` that doesn't get called by the test.

And I stole the `:caller` line from @akirak.

Closes #211 and #236.

jacktasia:
Thanks for opening this! Sorry for the delay in responding. I'll take a closer look at this tonight after work.

PR #309, opened 2019-11-13
Point byte-compiler to ivy-read and helm-make-source.

leungbk:
`declare-function` should be used to tell the byte-compiler where to look.

`helm-make-source` is preferred over `helm-build-sync-source` for packages with an optional Helm dependency using `declare-function` in this manner since `declare-function` doesn't work on macros.

I don't know much about Tramp, so I'm not sure if it's necessary to add in `(require 'tramp)` where the `tramp-` functions get invoked. Maybe I should add in the `require`s anyway, they shouldn't hurt.

At the time of writing, the `master` branch fails a test for me (related to my last PR) when I run `cask` followed by `make test-in-docker`, which was strange since it passed on Travis when I submitted. Locally, this new PR fails one additional test for me: it still expects `helm-build-sync-source`, even after I've changed the test accordingly. I hope the CI receives this PR kindly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,9 @@ test: unit
 unit:
 	${CASK} exec ert-runner
 
+compile:
+	${CASK} exec emacs -batch -L . -f batch-byte-compile dumb-jump.el
+
 install:
 	${CASK} install
 
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -1961,6 +1961,8 @@ This is the persistent action (\\[helm-execute-persistent-action]) for helm."
           (plist-get result :line)
           (s-trim (plist-get result :context))))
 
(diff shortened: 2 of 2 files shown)


leungbk:
Looks like I'll have to learn `el-mock` more carefully. The `require` statements now force `helm` to actually be loaded, which we don't want in the tests, but do want for normal use (since `helm` won't necessarily be loaded at the time `dumb-jump-go` is called).

And the Emacs 24 versions all seem to be incompatible with some feature or another of `helm` or `ivy`. So I'll have to figure out how to make the `require`s relevant only during normal usage.

jacktasia:
Thanks for opening this! Again, sorry for the delayed reply. Everything you mentioned makes sense to me to change if you're willing. Thanks again.

PR #330, opened 2020-04-03
Dropped f dependency

phikal:
Hi again,

during my last pull request I was reading through the code, and was a bit disappointed to see how much the f, s and dash libraries are bing used. _I personally_, consider it bad elisp style, and wanted to suggest dropping them incrementally for the sake of legibility, speed, and good style.

My issue is the following: If one takes a look at how many of the `f-...` functions are defined, they are often just aliases for existing and perfectly good Emacs functions (for example `f-exists?` and `file-exists-p). In other cases they promote an inefficient style that ignores some of the strengths of Elisp, such as using string over buffers (see my `dumb-jump-read-config` rewrite for example). 

If you're ok with my proposition to stick more to "vanilla" Elsip, I'll submit a few more push requests soon to remove the other two libraries and possibly rewrite a few functions to avoid unnecessary consing where possible. If not, it's no problem at all.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -3,7 +3,7 @@
 ;; Author: jack angers and contributors
 ;; Url: https://github.com/jacktasia/dumb-jump
 ;; Version: 0.5.3
-;; Package-Requires: ((emacs "24.3") (f "0.20.0") (s "1.11.0") (dash "2.9.0") (popup "0.5.3"))
+;; Package-Requires: ((emacs "24.3") (s "1.11.0") (dash "2.9.0") (popup "0.5.3"))
 ;; Keywords: programming
 
 ;; Dumb Jump is free software; you can redistribute it and/or modify it
@@ -28,7 +28,6 @@
 
 ;;; Code:
 (require 'etags)
-(require 'f)
 (require 's)
 (require 'dash)
 (require 'popup)
@@ -1895,7 +1894,7 @@ for user to select.  Filters PROJ path from files for display."
   "Keep looking at the parent dir of FILEPATH until a denoter file/dir is found."
   (s-chop-suffix
    "/"
-   (f-expand
+   (expand-file-name
     (or
      dumb-jump-project
      (locate-dominating-file filepath #'dumb-jump-get-config)
@@ -1905,10 +1904,10 @@ for user to select.  Filters PROJ path from files for display."
   "If a project denoter is in DIR then return it, otherwise
 nil. However, if DIR contains a `.dumbjumpignore' it returns nil
 to keep looking for another root."
-  (if (f-exists? (f-join dir ".dumbjumpignore"))
+  (if (file-exists-p (expand-file-name ".dumbjumpignore" dir))
       nil
     (car (--filter
-          (f-exists? (f-join dir it))
+          (file-exists-p (expand-file-name it dir))
           dumb-jump-project-denoters))))
 
 (defun dumb-jump-get-language (file)
@@ -1921,7 +1920,7 @@ to keep looking for another root."
                        (dumb-jump-get-mode-base-name))))
     (if (member language languages)
       language
-      (format ".%s file" (or (f-ext file) "")))))
+      (format ".%s file" (or (file-name-extension file) "")))))
 
 (defun dumb-jump-get-mode-base-name ()
   "Get the base name of the mode."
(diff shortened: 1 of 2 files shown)


phikal:
I'm reading the test results, and I have a few comments:

1. In `dumb-jump-cpp-test1`, the issue is that `dumb-jump-go` is returning the relative file name, as opposed to the absolute file name. I don't think this is strictly speaking a mistake, but I'll try to fix it.
2. In `dumb-jump-exclude-path-test` the files are returned in the reverse order which is trivial to fix, but then I don't understand why `/home/travis/dumb-jump/test/data/proj1/ignored2` is expected, when the file includes `/ignore2`, an absolute file path?

jacktasia:
> a bit disappointed to see how much the f, s and dash libraries are bing used. I personally, consider it bad elisp style, and wanted to suggest dropping them incrementally for the sake of legibility, speed, and good style.

I've definitely considered dropping these before because I'd much rather use built-ins then bring in a whole dependency, but in practice, this never worked very well particularly for supporting older versions of emacs. While some are merely better-named aliases for builtins some are doing quite a bit to ensure backward compatibility (if I remember correctly).

If you can get all the tests to pass without adding any complicated helper functions into the code base then I am definitely supportive and will merge. I just don't want to add complexity or duplicate (then need to separately maintain) existing code that already exists in these libraries.

jacktasia:
> I'm reading the test results, and I have a few comments:
> 
> 1. In `dumb-jump-cpp-test1`, the issue is that `dumb-jump-go` is returning the relative file name, as opposed to the absolute file name. I don't think this is strictly speaking a mistake, but I'll try to fix it.
> 2. In `dumb-jump-exclude-path-test` the files are returned in the reverse order which is trivial to fix, but then I don't understand why `/home/travis/dumb-jump/test/data/proj1/ignored2` is expected, when the file includes `/ignore2`, an absolute file path?

1. I can't remember the details but for the most part if I am looking for an absolute path it's for a good reason and to get around weirdness with projects that have duplicated path parts (multiple `/src`s etc)
2. The `-/ignored2` is not actually an absolute file path because searches are always relative to the directory/file they are explicitly searching. That is, if you try `ag`'s `--ignore-dir` option it will treat both ways the same, which if I remember correctly is why I added both versions.

phikal:
> The -/ignored2 is not actually an absolute file path because searches are always relative to the directory/file they are explicitly searching. That is, if you try ag's --ignore-dir option it will treat both ways the same, which if I remember correctly is why I added both versions.

Ok, but then `dumb-jump-include-path-test` shouldn't be expecting an absolute path (`/etc/var/some/code`). I've been jumping back and forth between fixing `dumb-jump-include-path-test` and `dumb-jump-exclude-path-blank-test`, because the behaviour of the one seems to contradict the other.

jacktasia:
@phikal Thanks for doing this! This is looking good. I am thinking of merging this soon. Do you have any additional concerns before I do?

phikal:
No, I have nothing more to add to this pull request. The only caveat would be to mention that I'm close to finishing a near 100% rewrite of the packet. As mention above, I started rewriting the code to drop s and dash, one thing lead to another and now I have a 5000+ line diff. I'm uncertain if submitting it as a pull request even makes sense, so if you're not too interested in that, merging this seems like a good first step.

jacktasia:
> No, I have nothing more to add to this pull request. The only caveat would be to mention that I'm close to finishing a near 100% rewrite of the packet. As mention above, I started rewriting the code to drop s and dash, one thing lead to another and now I have a 5000+ line diff. I'm uncertain if submitting it as a pull request even makes sense, so if you're not too interested in that, merging this seems like a good first step.

Wow, that's quite the change. Since you did the work you might as well open a PR. I'd be very curious to see what all changed, but to be perfectly honest I might not accept that drastic of a change.

That said, I'll keep testing this PR and probably merge this one soon. Thanks again!

phikal:
I'll open a PR when it's stable. Can't grantee when that will be though :/

---

In case you're interesting, here's a snapshot of the current state: https://0x0.st/iQE0.el. It doesn't work yet and I have to still rework `dumb-jump--result-follow` and `dumb-jump-process-results` and adapt few interactive commands, but the design should stay the same. The main additions to notice is the usage of EIEIO and that the search-process is started asynchronously.

PR #334, opened 2020-04-17
Rewrite

phikal:
As mentioned in #330, this is a suggestion for rewriting the internals of dumb-jump. The main changes are:

- Using EIEIO (Emacs' CLOS implementation) to implement common and differing behaviour between different searcher-tools. This is the class-hierarchy:  

   ```
  +--dumb-jump-searcher
       +--dumb-jump-grep
       |    +--dumb-jump-gnu-grep
       +--dumb-jump-git-grep
       |    +--dumb-jump-git-grep+ag
       +--dumb-jump-rg
       +--dumb-jump-ag
            +--dumb-jump-git-grep+ag
   ```

  Next to the methods, `dumb-jump-searcher` also has attributes for all the values that were previously manually passed from function call to function call.
- Dropping `s` and `dash`. Replaced with either elisp primitives with `cl-lib` functions. This also means changing the algorithms from a functional (`--filter`, `--map`, ...), to more imperative and where applicable destructive approach.
- Reordered functions into multiple sections (Utility Functions, Configuration, Query Constructor, Result Handling, Interactive Commands).
- Queries can be executed asynchronously. The variable `dumb-jump-async-p` controls this behaviour. By default it's activated, but could be changed for a xref backend.
- Regular expressions in `dumb-jump-find-rules` no longer have the `JJJ` and `\j` substrings, but now store them as a list with symbols.
- Assertions have been added where applicable. This could help replacing many of the white-box tests with black-box approaches.
- Avoided lines longer than 80 characters, both in code and docstrings.

A few features were dropped, but could be re-added:
- Custom ivy/helm backends
- Popups (I would re-add it as a weak-dependency but also support posframe for post Emacs 26)
- Max "fetch-time" warnings.

I also have to mention, that there are a few changes that would break backwards compatibility. These could be fixed by `:set` functions to the defcustom definitions, if maintaining them is a priority.

I know that this a lot, but I hope that it seems cleaner. I'm totally open to any criticism and change-requests. The tests would have to be re-written too, but I'm going to wait a bit before starting that. I hope that merging this patch could be a long-term goal, and know that I'm ready to help in any way necessary if you are interested in it.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -1994,7 +1994,7 @@ binary name.")
                  (when (and (not (delq (setq type (car res)) res))
                             type)
                    (push (list :type type) query)))))
-        (oset searcher regexps
+        (setf (oref searcher regexps)
               (or (dumb-jump-table-query-1 dumb-jump-find-rules
                                            query :regex)
                   (and dumb-jump-fallback-search
@@ -2027,7 +2027,8 @@ and then concatenated into a single string."
   "Check if ag is installed."
   (with-temp-buffer
     (shell-command (concat dumb-jump-ag-cmd " --version") t)
-    (oset searcher found-ag-p (and (search-forward "ag version" nil t) t))))
+    (setf (oref searcher found-ag-p)
+          (and (search-forward "ag version" nil t) t))))
 
 (cl-defmethod dumb-jump-check-usable ((searcher dumb-jump-ag))
   "Check if ag is installed."
@@ -2132,7 +2133,7 @@ and then concatenated into a single string."
   "Check if git grep is installed."
   (with-temp-buffer
     (shell-command (concat dumb-jump-git-cmd " grep") t)
-    (oset searcher found-git-grep-p
+    (setf (oref searcher found-git-grep-p)
           (and (search-forward "fatal: no pattern given" nil t) t))))
 
 (cl-defmethod dumb-jump-check-usable ((searcher dumb-jump-git-grep))
@@ -2193,9 +2194,8 @@ and then concatenated into a single string."
                                    (oref searcher root))
                  (oref searcher root))
                 files)))
-      (oset searcher ag-query (mapconcat #'identity
-                                         `("(" ,@files ")")
-                                         "|")))))
+      (setf (oref searcher ag-query)
+            (mapconcat #'identity `("(" ,@files ")") "|")))))
 
 
 ;;; General Grep Implementation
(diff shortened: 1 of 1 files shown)


phikal:
@jacktasia Any comments? If you're not interested at all, I'd rather close the pull request.

jacktasia:
Thanks for opening this! Sorry I haven't had a chance to take a closer look yet. I am definitely interested in taking a closer look at what you've done here, but obviously it's a super big change and things are a bit crazy.

I close this for now, but I'l add comments when I can dig in more.

phikal:
No problem, we're all experiencing the same right now, I just wanted to make sure the issue wasn't just buried in your inbox.

andschwa:
FWIW (I get life is freaking crazy right now) I would love to see this happen. I've added and removed the dumb-jump package a few different times since I primarily use xref and don't want to have to switch to different bindings when I'm in e.g. a Bash script (where dumb-jump comes in really handy), so this kind of rewrite to use built-in Emacs-isms and provide xref integration is very welcome! I was literally looking at this package thinking, _how hard would it be to make it provide a function to `xref-backend-functions`?_ so I'm glad I'm not the first person. I'm excited to see where this eventually lands! Thanks for all your work.

phikal:
I had a discussion on Reddit a few days ago, and the idea came up to propose a separate Pull request, just implementing xref from phikal@ab26b48, so to then "deprecate" the old UI and eventually replace it. I already have a sketch locally, but it's not quite working yet. If I have the time, I'm going to try and propose something in the next few days.

phikal:
I don't know what was wrong, but it's "working" now: https://github.com/phikal/dumb-jump/tree/xref-legacy. Probably hadn't evaluated everything properly. It isn't as accurate as `dumb-jump-go` as of writing, because the filtering and aggressive jumping is missing, but that shouldn't be too hard to add.

jacktasia:
> the idea came up to propose a separate Pull request, just implementing xref from phikal/dumb-jump@ab26b48,

Awesome, definitely very supportive of this. Thanks!

phikal:
@andschwa FYI, the pure xref interface has been merged.

phikal:
I recently found out that ELPA hosts [cl-generic](http://elpa.gnu.org/packages/cl-generic.html), that would enable this patch to be compatible with pre-Emacs 25 versions. I might propose a new patch series in the coming months to rework the code on a step-by-step basis (probably starting with the refactoring into classes and ending with the removal of dash and s).

PR #337, opened 2020-05-11
Failing tests for php

cweiske:
"dumb-jump-php-method-static" works locally, but not in docker (no idea why, sorry)

"dumb-jump-php-method-normal" searches for "$user->getSelf", but the symbol should be "getSelf" only. I do not know why it does not search for "getSelf" here.

### DIFF
--- a/test/data/proj4-php/User.php
+++ b/test/data/proj4-php/User.php
@@ -1,0 +1,35 @@
+<?php
+/**
+ * dumb-jump unit test class
+ *
+ * @property       string $firstname
+ * @property-read  string $lastname
+ * @property       int    $age
+ * @property-write User   $sibling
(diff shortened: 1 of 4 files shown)


cweiske:
(this PR is meant as a base for discussion)

jacktasia:
> "dumb-jump-php-method-normal" searches for "$user->getSelf", but the symbol should be "getSelf" only. I do not know why it does not search for "getSelf" here.

`(thing-at-point 'symbol)` is being used under the hood. if you're in `fundamental` (no) mode that will be `$user->getSelf` if it's in `php-mode` it will be `getSelf`

it might be enough to just do what's happening in `dumb-jump-get-lang-major-mode-test`

PR #348, opened 2020-06-26
Update README for xref interface

phikal:
Sorry for the delay, I thought it would be more work to update the README, but it ended up just being a few lines.

I took the liberty of removing references to the legacy commands, as I believe they shouldn't be promoted anymore, for the sake of simplicity. If you want, a note may be added on how to replace old features using xref.

When it came to rewriting the `use-package` section, I decided that it wasn't worth it, since enabling dumb-jump was already explained above. Hope this is ok.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,15 +87,16 @@ Dumb Jump performs best with The Silver Searcher `ag` ([ag install instructions]
 
 #### Basic
 
-Adding `(dumb-jump-mode)` to your `.emacs` will enable the key bindings for two interactive Dumb Jump functions:
+To enable the [xref][] backend, evaluate
 
-* `dumb-jump-go` <kbd>C-M-g</kbd> core functionality. Attempts to jump to the definition for the thing under point
-* `dumb-jump-back` <kbd>C-M-p</kbd> jumps back to where you were when you jumped. These are chained so if you go down a rabbit hole you can get back out or where you want to be.
-* `dumb-jump-quick-look` <kbd>C-M-q</kbd> like `dumb-jump-go` but **only** shows tooltip with `file`, `line`, and `context` it does not jump.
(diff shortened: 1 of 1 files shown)


jacktasia:
Thanks for doing this! I think we shouldn't remove anything yet just add a bold `obsolete` as either an extra header or next to each line. My thinking that as long as the code still exists then the documentation should too. Then (thanks to your other PR) both the code and documentation will be marked as obsolete.

I am happy to do this if you don't mind me pushing into your branch

phikal:

> Thanks for doing this! I think we shouldn't remove anything yet just
> add a bold `obsolete` as either an extra header or next to each
> line. My thinking that as long as the code still exists then the
> documentation should too. Then (thanks to your other PR) both the code
> and documentation will be marked as obsolete.

I think you're right that it should be documented, but from my own
experience of being lazy I'd be weary to have copy-paste'able code lying
around in the README, because even with a bold "obsolete" notice, people
will still yank into their config and then get annoyed by the "obsoleted"
notices.

> I am happy to do this if you don't mind me pushing into your branch

Change whatever you think is right.

-- 
	Philip K.

phikal:
The ".emacs" section should also be updated to note what variables don't have any effect anymore, when using xref.

phikal:
The last version should document the legacy API properly, without creating the inclination to just copy-paste the first snippet one finds in the README. I've also removed all options from the ".emacs" section that don't have any effect any more, but they're still listed in the legacy section.

Hope this is better.

phikal:
I was just re-reading the README, and noticed that the [Hydra](https://github.com/jacktasia/dumb-jump#hydra-for-effieciency) still uses legacy commands. Should this be updated? Does it still make sense? (I don't use Hydra, so I can't say).

PR #349, opened 2020-06-26
Mark legacy commands as obsolete

phikal:
Something along these lines was originally implemented in #343, but I decided to leave it out since it's a different issue.

Marking the commands as obsolete doesn't break anything, so backwards compatibility is ensured. Furthermore, this will only take effect if xref is installed, so the theoretical Emacs 24.1 user won't be annoyed ^^.

---

If you're not interested in obsoleting the old commands, _please_ feel free to say so. If that's the case, I continue developing the changes from #334 into it's own fork, since it has basically been rewritten 100%, dropping all backwards compatibility so that it may be merged into ELPA.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -2906,6 +2906,25 @@ Using ag to search only the files found via git-grep literal symbol search."
 
 ;;; Xref Backend
 (when (featurep 'xref)
+  (dolist (obsolete
+	   '(dumb-jump-mode
+	     dumb-jump-go
+	     dumb-jump-go-prefer-external-other-window
+	     dumb-jump-go-prompt
+	     dumb-jump-quick-look
+	     dumb-jump-go-other-window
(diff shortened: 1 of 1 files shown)


jacktasia:
Thanks!

amno1:
I would definitely like to keep them not obsolete and to have them as interface into dumb-jump so I don't need to rebind my keys and what not.

phikal:
Arthur Miller <notifications@github.com> writes:

> I would definitely like to keep them not obsolete and to have them as
> interface into dumb-jump so I don't need to rebind my keys and what
> not.

The whole point is that you don't have to rebind anything. Xref's "go to
definition" is already bound by default (M-. is bound to the command
xref-find-definitions). All you need to do is register the backend, but
as the README explains, that's a single line of elisp.

-- 
	Philip K.

PR #350, opened 2020-07-15
Updated Commentary section to explain xref

phikal:
Updating the commentary section with how to use the xref interface should make it easier for older users to switch and newer users to configure everything.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -21,6 +21,10 @@
 
 ;;; Commentary:
 
+;; Dumb Jump is an Emacs "jump to definition" package with support for 40+ programming languages that favors
+;; "just working" over speed or accuracy.  This means minimal -- and ideally zero -- configuration with absolutely
+;; no stored indexes (TAGS) or persistent background processes.
+;;
 ;; Dumb Jump provides a xref-based interface for jumping to
 ;; definitions. It is based on tools such as grep, the silver searcher
 ;; (https://geoff.greer.fm/ag/), ripgrep
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -21,10 +21,22 @@
 
 ;;; Commentary:
 
-;; Dumb Jump is an Emacs "jump to definition" package with support for 40+ programming languages that favors
-;; "just working" over speed or accuracy.  This means minimal -- and ideally zero -- configuration with absolutely
-;; no stored indexes (TAGS) or persistent background processes.  Dumb Jump performs best with The Silver Searcher
-;; `ag` or ripgrep `rg` installed.  Dumb Jump requires at least GNU Emacs 24.3.
+;; Dumb Jump provides a xref-based interface for jumping to
+;; definitions. It is based on tools such as grep, the silver searcher
+;; (https://geoff.greer.fm/ag/), ripgrep
+;; (https://github.com/BurntSushi/ripgrep) or git-grep
+;; (https://git-scm.com/docs/git-grep). This means minimal -- and
+;; ideally zero -- configuration with absolutely no stored indexes
+;; (TAGS) or persistent background processes.
+;;
+;; To enable Dumb Jump, add the following to your initialisation file:
+;;
+;;    (add-hook 'xref-backend-functions #'dumb-jump-xref-activate)
+;;
+;; Now pressing M-. on an identifier should open a buffer at the place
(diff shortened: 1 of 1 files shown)


dumb-jump.el
```diff
@@ -23,8 +23,24 @@
...
-;; `ag` or ripgrep `rg` installed.  Dumb Jump requires at least GNU Emacs 24.3.
+;; no stored indexes (TAGS) or persistent background processes.
+;;
+;; Dumb Jump provides a xref-based interface for jumping to
+;; definitions. It is based on tools such as grep, the silver searcher
+;; (https://geoff.greer.fm/ag/), ripgrep
+;; (https://github.com/BurntSushi/ripgrep) or git-grep
+;; (https://git-scm.com/docs/git-grep). This means minimal -- and
+;; ideally zero -- configuration with absolutely no stored indexes
+;; (TAGS) or persistent background processes.
```
phikal, line 34: The last two 1/2 lines are now duplicated.

dumb-jump.el
```diff
@@ -23,8 +23,24 @@
 
 ;; Dumb Jump is an Emacs "jump to definition" package with support for 40+ programming languages that favors
 ;; "just working" over speed or accuracy.  This means minimal -- and ideally zero -- configuration with absolutely
-;; no stored indexes (TAGS) or persistent background processes.  Dumb Jump performs best with The Silver Searcher
-;; `ag` or ripgrep `rg` installed.  Dumb Jump requires at least GNU Emacs 24.3.
+;; no stored indexes (TAGS) or persistent background processes.
```
phikal, line 26: Should these lines be folded, so that the whole section has the same width?

TonyRode:
The `(add-hook 'xref-backend-functions #'dumb-jump-xref-activate)` in this pull-request works fine whereas the `(add-to-list 'xref-backend-functions #'dumb-jump-xref-activate)` from the README.md didn't.

phikal:
@TonyRode Did you evaluate `(add-to-list 'xref-backend-functions #'dumb-jump-xref-activate)` in a major-mode hook or "globally" in your initialisation file?

TonyRode:
> @TonyRode Did you evaluate `(add-to-list 'xref-backend-functions #'dumb-jump-xref-activate)` in a major-mode hook or "globally" in your initialisation file?

I tried an `eval-region` on that line and also added it in my ~/.emacs followed by a `M-x load-file RET ~/.emacs` or even quitting / starting Emacs. 
At starting, error message was `Symbol's value as variable is void: xref-backend-functions`.
I can give you the full error backtrace if needed.

phikal:
It seems that one of the differences between `add-hook` and `add-to-list` is that `add-hook` binds the hook if it hasn't been defined yet -- for example before a package has been loaded. From the docstring:

> If HOOK is void, it is first set to nil.

So it's probably better to change that in the README.

Just one more question, does anything change if you wrap the `add-to-list` in 

~~~el
(with-eval-after-load 'dumb-jump
  ...)
~~~

?

TonyRode:
You mean, like this ?

`
(with-eval-after-load 'dumb-jump
  (add-to-list 'xref-backend-functions #'dumb-jump-xref-activate))
`

No error at loading ~/.emacs but the dumb-jump doesn't seem to be loaded, because when I then hit M-. on a word I'm asked to provide a TAGS table.

phikal:
You're right, my mistake. I forgot that `dumb-jump-xref-activate` is the autoloaded function, so it can't load itself -- but that's not the issue, xref is probably not loaded, and therefore hasn't defined `xref-backend-functions`. Maybe this would be better?

~~~el
(with-eval-after-load 'xref
 (add-to-list 'xref-backend-functions #'dumb-jump-xref-activate))
~~~

But since that would be more cumbersome, I agree that `add-hook` should be preferred (#358 proposes the change).

TonyRode:
> You're right, my mistake. I forgot that `dumb-jump-xref-activate` is the autoloaded function, so it can't load itself -- but that's not the issue, xref is probably not loaded, and therefore hasn't defined `xref-backend-functions`. Maybe this would be better?
> 
> ```emacs-lisp
> (with-eval-after-load 'xref
>  (add-to-list 'xref-backend-functions #'dumb-jump-xref-activate))
> ```
> 
> But since that would be more cumbersome, I agree that `add-hook` should be preferred (#358 proposes the change).

Yes this works :+1:

Issue #345, opened 2020-06-26
New byte compiler warnings


thisch:
When updating dumb-jump to the latest version, I noticed the following byte compiler-warnings:

```
dumb-jump.el:74:1: Warning: defcustom for
    ‘dumb-jump-ivy-jump-to-selected-function’ fails to specify type
dumb-jump.el:74:1: Warning: defcustom for
    ‘dumb-jump-ivy-jump-to-selected-function’ fails to specify type
dumb-jump.el:2053:1: Warning: Unused lexical argument ‘config’

In dumb-jump-handle-results:
dumb-jump.el:2270:153: Warning: reference to free variable ‘rel-cur-file’
dumb-jump.el:2279:1: Warning: Unused lexical argument ‘use-tooltip’
dumb-jump.el:2279:1: Warning: Unused lexical argument ‘look-for’
dumb-jump.el:2390:1: Warning: Unused lexical variable ‘modified-file-buffers’

In dumb-jump-goto-file-line:
dumb-jump.el:2449:17: Warning: ‘find-tag-marker-ring’ is an obsolete variable
    (as of 25.1); use ‘xref-push-marker-stack’ or ‘xref-pop-marker-stack’
    instead.
dumb-jump.el:2765:1: Warning: Unused lexical argument ‘lang’
dumb-jump.el:2785:1: Warning: Unused lexical argument ‘cur-file’
dumb-jump.el:2837:1: Warning: Unused lexical argument ‘exclude-paths’
dumb-jump.el:2837:1: Warning: Unused lexical argument ‘lang’
dumb-jump.el:2942:13: Warning: reference to free variable ‘ctx-type’
dumb-jump.el:2942:93: Warning: reference to free variable ‘prefer-external’
dumb-jump.el:2942:140: Warning: reference to free variable ‘cur-file’
dumb-jump.el:2942:149: Warning: reference to free variable ‘rel-cur-file’
```

PR #354, opened 2020-07-18
Fix byte-compiler warnings in usage of debug macro

thisch:
This fixes three "reference to free variable" warnings.

partially fixes https://github.com/jacktasia/dumb-jump/issues/345.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -3030,14 +3030,13 @@ Using ag to search only the files found via git-grep literal symbol search."
 
       (dumb-jump-debug-message
        look-for
-       ctx-type
+       (plist-get info :ctx-type)
        var-to-jump
        (pp-to-string match-cur-file-front)
        (pp-to-string results)
-       prefer-external
        match-cur-file-front
        proj-root
-       cur-file)
+       (plist-get info :file))
       (cond ((eq issue 'nogrep)
              (dumb-jump-message "Please install ag, rg, git grep or grep!"))
             ((eq issue 'nosymbol)


phikal:
These values are actually accessible, just not bound (eg. `ctx-type` can be retrieved using `(plist-get info :ctx-type)`).

thisch:
> These values are actually accessible, just not bound (eg. `ctx-type` can be retrieved using `(plist-get info :ctx-type)`).

So do you suggest that I use `(plist-get info ...)` instead?

phikal:
Thomas Hisch <notifications@github.com> writes:

>> These values are actually accessible, just not bound (eg. `ctx-type` can be retrieved using `(plist-get info :ctx-type)`).
>
> So do you suggest that I use `(plist-get info ...)` instead? 

My point is that the values are still accessible, just the variables are
missing, yes, because there's not point why the debugging information
should be decreased because of that fact.

-- 
	Philip K.

thisch:
> Thomas Hisch <notifications@github.com> writes:
> > These values are actually accessible, just not bound (eg. `ctx-type` can be retrieved using `(plist-get info :ctx-type)`). So do you suggest that I use `(plist-get info ...)` instead?
> My point is that the values are still accessible, just the variables are missing, yes, because there's not point why the debugging information should be decreased because of that fact.
> […](#)
> -- Philip K.

I uploaded a new version of the PR. AFAICS `prefer-external` can not be determined from inside `xref-backend-definitions`.

phikal:
Yes, the xref interface doesn't (as of now) differentiate between "external" and "internal" matches, so dropping it for now is probably the best.

thisch:
Good, then we can merge this PR right?

phikal:
I won't object, for whatever it's worth.

jacktasia:
Thanks!

PR #356, opened 2020-07-28
Implements xref-backend-apropos

nhojb:
This permits jumping to a custom symbol via xref-find-apropos

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -3039,8 +3039,8 @@ Using ag to search only the files found via git-grep literal symbol search."
 			   (list var-to-jump)
              match-cur-file-front))))))
 
-    (cl-defmethod xref-backend-apropos ((_backend (eql dumb-jump)) symbol)
-      (xref-backend-definitions 'dumb-jump symbol)))
+    (cl-defmethod xref-backend-apropos ((_backend (eql dumb-jump)) pattern)
+      (xref-backend-definitions 'dumb-jump pattern)))
 
 ;;;###autoload
 (defun dumb-jump-xref-activate ()
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -3037,7 +3037,10 @@ Using ag to search only the files found via git-grep literal symbol search."
 			   0)))
 		       (if do-var-jump
 			   (list var-to-jump)
-			 match-cur-file-front)))))))
+             match-cur-file-front))))))
+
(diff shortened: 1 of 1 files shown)


dumb-jump.el
```diff
@@ -3037,7 +3037,10 @@ Using ag to search only the files found via git-grep literal symbol search."
 			   0)))
 		       (if do-var-jump
 			   (list var-to-jump)
-			 match-cur-file-front)))))))
+             match-cur-file-front))))))
+
+    (cl-defmethod xref-backend-apropos ((_backend (eql dumb-jump)) symbol)
```
phikal, line 3042: The generic method names the second parameter `pattern` (some backends choose `regexp`). I think it would be use a term along these lines to make clear that the argument isn't generated by `xref-backend-identifier-at-point`, and therefore has no context.

phikal:
How accurate have the results been in your testing? I've tried it in two different C repositories, and it basically boils down to a grep (which isn't surprising).

I'm not sure if this should be considered a bug, but in some cases, comments are also included in the match results, polluting the actually interesting results. I can't see an easy way to solve this though, because in my case they are all part of multi-line comments.

dumb-jump.el
nhojb: Done

nhojb:
I implemented this as a (xref) replacement for `dumb-jump-go-prompt` - which now gives a deprecation warning.

I had assumed that that matches would be identical to the latter.

phikal:
> I implemented this as a (xref) replacement for dumb-jump-go-prompt - which now gives a deprecation warning.

That should be `C-u M-.` (I realized there's an issue here too), if you want to find a specific symbol (eg. `dumb-jump-concat-command` and only `dumb-jump-concat-command`). `xref-find-apropos` is meant for finding patterns via regular expressions (eg. `dumb-jump-.*language`).

That being said, I still think it's a good addition!

nhojb:
I don't seem to have anything bound to `C-u M-.` (default xref setup).

I found that `M-.` or `M-?` will result in an error (`apply: Wrong type argument: number-or-marker-p, nil`) if there is no symbol at the current point. These commands work ok in an elisp file, but not in a Swift file (for example).

I see you've also noticed this now :-)

That's why I tried `C-M-.` (xref-find-apropos) which did seem to work fine (once I'd implemented `xref-backend-apropos`).

phikal:
FYI `C-u` a prefix argument/modifier for the standard `M-.` (`xref-find-definitions`) command, it's not directly bound.

PR #359, opened 2020-07-29
Whitespace cleanup

phikal:
a4aece073ecf3e7b16a24ce84d47b8e5b98fa558 pointed out that there are a few trailing whitespaces in the codebase, so to avoid issues like these, this patch should remove all current trailing whitespaces + convert all tabs to whitespaces, as recommended in the [Elisp Style Guide](https://github.com/bbatsov/emacs-lisp-style-guide#source-code-layout--organization). To enforce this in the future, a `.dir-locals.el` file has been added to override local user preferences.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,0 +1,5 @@
+;;; Directory Local Variables
+;;; For more information see (info "(emacs) Directory Variables")
+
+((emacs-lisp-mode
+  (indent-tabs-mode . nil)))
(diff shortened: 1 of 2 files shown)


jacktasia:
@phikal Thanks for doing this! Out of curiosity did you use a tool for this or did you do it by hand? I gave up looking for a `gofmt`/`black` style formatter  for emacs lisp a while ago but curious if one is out there.

phikal:
No, I just ran `indent-region`, `whitespace-cleanup` and to be safe `untabify` over the whole buffer.

phikal:
I also cleaned up `dumb-jump-debug-message` a bit, I hope it's ok to add that to this pull request.

PR #366, opened 2020-09-12
Add support for TCL and fix some bugs

CeleritasCelery:
This PR adds support for TCL and fixes some bugs I found along the way. I can split it into multiple PR's if needed, but my TCL code does not work without the bug fixes here.
### bugs: 
- getting the `:left` and `:right` of a symbol does not work correctly when the symbol is at the start of the line.
You can see this with 
```lisp
(dumb-jump-get-point-context "foo bar baz" "foo" 0) => (:left "f" :right "bar baz")
```
notice that the left side is "f", which is not correct. I also took the chance to simplify the code and update the tests.

- Git grep does not work if not file extensions are specified 

Currently if no file extensions are specified then git grep will not be given a path and will therefore only search from the current directory instead of the root. So in that case we add the magic path `:/` to search from the project root. Note that this only works in Git versions greater then 1.7.6.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -47,6 +47,7 @@
 (require 'dash)
 (require 'popup)
 (require 'cl-generic nil :noerror)
+(require 'cl-lib)
 
 (defgroup dumb-jump nil
   "Easily jump to project function and variable definitions"
@@ -1776,18 +1777,6 @@ inaccurate jump).  If nil, jump without confirmation but print a warning."
         (setq dumb-jump--grep-installed? variant))
     dumb-jump--grep-installed?))
 
-(defun dumb-jump-find-start-pos (line-in look-for cur-pos)
-  "Find start column position for LINE-IN of LOOK-FOR using CUR-POS as a hint."
-  (let ((is-found nil)
-        (line (s-replace "\t" (s-repeat tab-width " ") line-in)))
-    (while (and (> cur-pos 0) (not is-found))
-      (let* ((char (substring line cur-pos (1+ cur-pos)))
-             (is-at (s-index-of char look-for)))
-        (if (null is-at)
-            (setq is-found t)
-          (setq cur-pos (1- cur-pos)))))
-    (1+ cur-pos)))
-
 (defun dumb-jump-run-test (test cmd)
   "Use TEST as the standard input for the CMD."
   (with-temp-buffer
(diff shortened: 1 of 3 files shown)


dumb-jump.el
```diff
@@ -2111,6 +2108,13 @@ to keep looking for another root."
         (thing-at-point 'symbol)
       (thing-at-point 'symbol t))))
 
+(defun dumb-jump--get-symbol-start ()
+  "Get bounds of symbol at point"
+  (- (if (region-active-p)
+         (region-beginning)
```
phikal, line 2114: I'm not sure if using the beginning of the region is the expected behaviour, especially for xref.

dumb-jump.el
```diff
@@ -2111,6 +2108,13 @@ to keep looking for another root."
...
+       (car (bounds-of-thing-at-point 'symbol)))
+     (point-at-bol)))
```
phikal, line 2116: `point-at-bol` is a XEmacs compatibility function, I think it would be better to use `line-beginning-position` instead.

dumb-jump.el
```diff
@@ -1918,16 +1922,9 @@ Optionally pass t for RUN-NOT-TESTS to see a list of all failed rules"
...
-         (func-len (length func))
-         (sen-len (length line))
-         (right-loc-start (+ loc func-len))
-         (right-loc-end (length line))
-         (left (substring line 0 loc))
-         (right (if (> right-loc-end sen-len)
-                    ""
-                  (substring line right-loc-start right-loc-end))))
-    `(:left ,left :right ,right)))
+  (let* ((loc (or (cl-search func line :start2 cur-pos) 0)))
```
phikal, line 1925: Currently, we only attempt to load `cl-generic`, but not `cl-lib` or `cl-seq`. Since we're already using a few cl-lib functions, we should probably require `cl-lib` anyway.

dumb-jump.el
CeleritasCelery: `dumb-jump-get-point-symbol` returns the region (if it is active) for both normal and xref mode. So we want to model the same behavior.

test/dumb-jump-test.el
```diff
@@ -424,14 +424,14 @@
 (ert-deftest dumb-jump-context-point-test ()
   (let* ((sentence "mainWindow.loadUrl('file://')")
          (func "loadUrl")
-         (ctx (dumb-jump-get-point-context sentence func 15)))
+         (ctx (dumb-jump-get-point-context sentence func 11)))
```
jacktasia, line 427: I am not sure I understand why it's necessary to change `cur-pos`. It seems to imply that it won't now work _anywhere_ on the current symbol, or am I missing something here?

jacktasia:
Thanks for opening this! I just had a small question above

dumb-jump.el
phikal: You're right, I forgot that `dumb-jump-get-point-symbol` used `region-active-p`.

test/dumb-jump-test.el
CeleritasCelery: You are correct that it won't work anywhere in the current symbol. But the code will never place it inside the symbol. The `look-for` symbol comes from the code below

https://github.com/jacktasia/dumb-jump/blob/b47774bc167fbd5d336700a17a66f8bc2b7cec46/dumb-jump.el#L2139-L2140

And we know that the `cur-pos` hint will always be the start of the `dumb-jump-get-point-symbol` So there is no way for `cur-pos` to be anywhere in the symbol, it will always be the left of it. This was true even with the old code, but the old `dumb-jump-get-point-context` could handle the case where the hint was inside the symbol even though that case couldn't happen.

jacktasia (approved):
Thanks!

PR #368, opened 2020-09-12
Fix C++ function rule with colon in parameter list

jrosdahl:
A C++ function definition like

```c++
void MyClass::test(std::string) {}
```

is not found by dumb-jump. Fix this by allowing colon in parameter lists.

By the way, many thanks for dumb-jump!


jacktasia (approved):
Thanks!

PR #369, opened 2020-09-13
Fixed defcustom warnings.

jcs090218:
as title.


phikal:
Wouldn't it be better to specify the type as `function`, instead of `symbol`?

jcs090218:
Fixed!

jacktasia (approved):
Thanks!

PR #372, opened 2020-10-15
Always expand relative paths in xref backend

### LINKED ISSUE #333: relative paths are not handled correctly when the searcher is `git grep`
The format of the paths in the output of `git grep` is relative to project root, not absolute. Dumb Jump attempts to navigate to `directory_of_current_buffer/path/relative/to/project/root`, even though the grep had indeed identified the correct path.

A user can bypass this with `(setq dumb-jump-force-searcher 'ag)`, which is recommended anyway, so it's probably not common for people to encounter this issue. That fix works because the output of `ag` is an absolute path.

nickg:
I also ran into issue #333 when using `git grep` and the `xref` backend. The problem is when you select a result with a relative path from the `*xref*` buffer the file name is expanded relative to the `default-directory` in the `*xref*` buffer which may be different to the `default-directory` of the current file, which the `git grep` results are relative to. The `*xref*` buffer is reused between searches and its `default-directory` is never updated, so you typically hit this bug by doing a search from one file, which works, and then doing another search from a second file in a different directory.

Simple fix here which works for me is to always expand the search results to absolute path names.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -2668,16 +2668,10 @@ searcher symbol."
       nil)
      ((and parts line-num-raw)
       (if (= (length parts) 2)
-          (list (let ((path (expand-file-name (nth 0 parts))))
-                  (if (file-name-absolute-p (nth 0 parts))
-                      path
-                    (file-relative-name path)))
+          (list (expand-file-name (nth 0 parts))
                 (nth 1 line-num-raw) (nth 1 parts))
                                         ; this case is when they are searching a particular file...
-        (list (let ((path (expand-file-name cur-file)))
-                (if (file-name-absolute-p cur-file)
-                    path
-                  (file-relative-name path)))
+        (list (expand-file-name cur-file)
               (nth 1 line-num-raw) (nth 0 parts)))))))
 
 (defun dumb-jump-parse-response-lines (parsed cur-file cur-line-num)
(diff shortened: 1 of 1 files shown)


phikal:
You have to log into CircleCI, to see the issue, so I'll report it here.

These tests have failed:

-   FAILED  dumb-jump-ag-parse-test
-   FAILED  dumb-jump-git-grep-parse-test
-   FAILED  dumb-jump-grep-parse-no-filter-test
-   FAILED  dumb-jump-grep-parse-test
-   FAILED  dumb-jump-parse-response-line-test
-   FAILED  dumb-jump-rg-parse-test

And they all seem to have triggered the same kind of error:

```
Test dumb-jump-ag-parse-test condition:

    (ert-test-failed
     ((should
       (=
	(length parsed)
	2))
      :form
      (= 3 2)
      :value nil))
```

Triggered by tests like [this one](https://github.com/jacktasia/dumb-jump/blob/master/test/dumb-jump-test.el#L266).

nickg:
@phikal thanks, looks like the tests were relying on the existing relative path behaviour. I just pushed a simpler fix which only expands the paths in the xref backend, although this is probably not the same as the original issue #333.

phikal:
Xref is currently not tested, so I'm not sure if this breaks anything with any specific searcher (I'm inclined to say no).

jacktasia (approved):
Thanks!

PR #374, opened 2020-11-03
Don't raise an error when no symbol is found at a point

atodekangae:
- Current behavior: when `M-.` is pressed, an error occurs if there is no symbol at a point
- Expected behavior: a prompt will open, asking for an identifier

Originally this problem was fixed with the PR #357, but a regression has been introduced later with the PR #366. In addition to reverting the change in `xref-backend-identifier-at-point`, defining `xref-backend-identifier-completion-table` was required to fix all the errors, hence the diff in the commit.

### Notes on the definition of xref-backend-identifier-completion-table
Personally I define `xref-backend-identifier-completion-table` as follows (slightly modified from [this code](https://github.com/NicolasPetton/xref-js2/blob/6f1ed5dae0c2485416196a51f2fa92f32e4b8262/xref-js2.el#L125)):

```elisp
(cl-defmethod xref-backend-identifier-completion-table ((_backend (eql dumb-jump)))
  (let (words)
    (save-excursion
      (save-restriction
        (widen)
        (goto-char (point-min))
        (while (re-search-forward "[a-zA-Z_][0-9a-zA-Z_]*" nil t)
          (add-to-list 'words (match-string-no-properties 0)))
        (seq-uniq words)))))
```

This assumes that identifiers match the regular expression `[a-zA-Z_][0-9a-zA-Z_]*`, which is the case in programming languages I deal with frequently. However since in some programming languages (most notably those languages falling in the Lisp family) `-` (hyphen) can also be used in an identifier and I thought it would be inappropriate to limit the scope to a fraction of the languages this library supports, I defined it to always return `nil` in this PR.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -3006,11 +3006,13 @@ Using ag to search only the files found via git-grep literal symbol search."
                  "2020-06-26")
 
   (cl-defmethod xref-backend-identifier-at-point ((_backend (eql dumb-jump)))
-    (let ((start (dumb-jump--get-symbol-start)))
-      (and start (let* ((ident (dumb-jump-get-point-symbol))
-                        (line (dumb-jump-get-point-line))
-                        (ctx (dumb-jump-get-point-context line ident start)))
-                   (propertize ident :dumb-jump-ctx ctx)))))
+    (let ((bounds (bounds-of-thing-at-point 'symbol)))
+      (and bounds (let* ((ident (dumb-jump-get-point-symbol))
+			             (start (car bounds))
+			             (col (- start (point-at-bol)))
+			             (line (dumb-jump-get-point-line))
+			             (ctx (dumb-jump-get-point-context line ident col)))
+		            (propertize ident :dumb-jump-ctx ctx)))))
 
   (cl-defmethod xref-backend-definitions ((_backend (eql dumb-jump)) prompt)
     (let* ((info (dumb-jump-get-results prompt))
@@ -3061,7 +3063,10 @@ Using ag to search only the files found via git-grep literal symbol search."
                          match-cur-file-front))))))
 
   (cl-defmethod xref-backend-apropos ((_backend (eql dumb-jump)) pattern)
-    (xref-backend-definitions 'dumb-jump pattern)))
+    (xref-backend-definitions 'dumb-jump pattern))
+
+  (cl-defmethod xref-backend-identifier-completion-table ((_backend (eql dumb-jump)))
+    nil))
 
 ;;;###autoload
 (defun dumb-jump-xref-activate ()


phikal:
There seem to have been some wierd non-Emacs related errors in the CI: 

```
Importing package-keyring.gpg...
Importing package-keyring.gpg...done
Contacting host: melpa.org:443
gnutls.c: [0] (Emacs) Received alert:  Handshake failed
gnutls.el: (err=[-12] A TLS fatal alert has been received.) boot: (:priority NORMAL :hostname melpa.org :loglevel 0 :min-prime-bits 256 :trustfiles (/etc/ssl/certs/ca-certificates.crt) :crlfiles nil :keylist nil :verify-flags nil :verify-error nil :callbacks nil)
Failed to download `melpa' archive.
Contacting host: elpa.gnu.org:443
Package refresh done
Failed to download `gnu' archive.
Setting `package-selected-packages' temporarily since "emacs -q" would overwrite customizations
Setting `package-selected-packages' temporarily since "emacs -q" would overwrite customizations
Package `s-' is unavailable
```

> Personally I define xref-backend-identifier-completion-table as follows (slightly modified from this code):

Have you tried using syntax-table functions? Usually, every major mode should define what a symbol is, and I can imagine that that could be used to generate a list of symbols. The only issue is that the approach is limited and/or inefficient: Not all symbols are symbols you would want to look up (eg. type names, keywords, etc.), and not all symbols are in the current file. An alternative approach would be to use the regular expressions used for searching for symbols to find all instances of definitions. I could imagine that if you replaced the "JJJ" with `\(\_<.+?\_>\)` you would match most definitions, and maybe even more.

atodekangae:
> Have you tried using syntax-table functions?

I was not very familiar with syntax tables at the time of the previous comment, but some digging into the source code revealed that they are under the hood in the mechanics of this library for extracting a symbol at a point (invoked via `forward-thing 'symbol`):

```elisp
(defun forward-symbol (arg)
  "Move point to the next position that is the end of a symbol.
A symbol is any sequence of characters that are in either the
word constituent or symbol constituent syntax class.
With prefix argument ARG, do it ARG times if positive, or move
backwards ARG times if negative."
  (interactive "^p")
  (if (natnump arg)
      (re-search-forward "\\(\\sw\\|\\s_\\)+" nil 'move arg)
    (while (< arg 0)
      (if (re-search-backward "\\(\\sw\\|\\s_\\)+" nil 'move)
	  (skip-syntax-backward "w_"))
      (setq arg (1+ arg)))))
```

The regular expressions here search for sequences each of which consists of word-characters and symbols according to the current syntax table.

It seems to me that it would be a good idea to create a list of symbols the same way `dumb-jump` extracts a symbol at a point, namely using `forward-thing 'symbol`. At least the behavior would be consistent.

Here is one option that employs `forward-thing 'symbol`:

```elisp
(cl-defmethod xref-backend-identifier-completion-table ((_backend (eql dumb-jump)))
  (let (idents)
    (save-excursion
      (save-restriction
        (widen)
        (goto-char (point-min))
        (while (not (eobp))
          (forward-thing 'symbol 1)
          (let ((end-of-id (point)))
            (save-excursion
              (forward-thing 'symbol -1)
              (let ((ident (buffer-substring (point) end-of-id)))
                (add-to-list 'idents ident)))))
        (seq-uniq idents)))))
```

phikal:
That is better, but from experience I know that a 

```elisp
(whine (not (eobp)
  (forward-*)
  ...)
```

block is often slower than expected. Compare that to 

```elisp
(let (idents)
    (save-excursion
      (save-restriction
	(widen)
	(goto-char (point-min))
	(while (re-search-forward "\\(\\_<\\(?:\\s_\\|\\sw\\)+?\\_>\\)" nil t)
	  (push (match-string-no-properties 1) idents))
	(delete-dups idents))))
```

Which uses the search engine and the resulting search data directly. On my system, this appears to speed things up by a factor of 10x, but I haven't tested it extensivly yet.

jacktasia:
Sorry, still unclear to me why CircleCI suddenly stopped working. A bunch of errors that seem related to network weirdness...I guess I am going to have to change to start using GitHub Actions or something like that.

phikal:
> I could imagine that if you replaced the "JJJ" with \(\_<.+?\_>\) you would match most definitions, and maybe even more.

I've tried this, but the problem seems translating the regular expressions made for grep and co. to elisp regular expressions. My current attempt look like this, but it sadly returns an empty table:

```elisp
(cl-defmethod xref-backend-identifier-completion-table ((_backend (eql dumb-jump)))
  (let* ((cur-file (or (buffer-file-name) ""))
         (proj-root (dumb-jump-get-project-root cur-file))
         (proj-config (dumb-jump-get-config proj-root))
         (config (when (s-ends-with? ".dumbjump" proj-config)
                   (dumb-jump-read-config proj-root proj-config)))
         (lang (or (plist-get config :language)
                   (dumb-jump-get-language cur-file)))
         (gen-funcs (dumb-jump-pick-grep-variant proj-root))
         (searcher (plist-get gen-funcs :searcher))
         (regexes (dumb-jump-get-contextual-regexes lang nil searcher))
         (regex
          (format "\\(?:%s\\)" (mapconcat (lambda (regex)
                                            (replace-regexp-in-string
                                             "JJJ" "\\(?1:\\_<\\(?:\\s_\\)+?\\_>\\)"
                                             (replace-regexp-in-string
                                              "\\\\\\([{}()]\\)" "\1" regex)
                                             nil t))
                                          regexes "\\|")))
         idents)
    (save-excursion
      (save-restriction
	(widen)
	(goto-char (point-min))
	(while (re-search-forward regex nil t)
	  (push (match-string-no-properties 1) idents))
	(delete-dups idents)))))
```

If dumb-jump is ever to be rewritten internally, replacing the regular-expression syntax with something like like `rx` would probably help a lot.

jacktasia:
I moved things over to use GitHub Actions last night so tests are running again. This LGTM, but going to lean on @phikal for the final approval if he doesn't mind. Thanks everyone!

phikal (approved):
It's all fine by me :+1:

Issue #380, opened 2020-12-03
Obsolete customization please


Ergus:
Hi:
I really like dumb-jump, but I use tramp a lot, and in that case it doesn't work very well and is slower; so I use dumb-jump only in local buffers throw the legacy interface... but the obsolete warning is very annoying. 

I have a config like this:

```
(use-package dumb-jump
  :bind-keymap ("C-c j" . dumb-jump-mode-map)
  :init
  (which-key-add-key-based-replacements "C-c j" "dumb-jump")
  :custom
  (dumb-jump-selector 'ivy)
  :config
  (define-key dumb-jump-mode-map "j" #'dumb-jump-go)
  (define-key dumb-jump-mode-map "o" #'dumb-jump-go-other-window)
  (define-key dumb-jump-mode-map "e" #'dumb-jump-go-prefer-external)
  (define-key dumb-jump-mode-map "4" #'dumb-jump-go-prefer-external-other-window)
  (define-key dumb-jump-mode-map "i" #'dumb-jump-go-prompt)
  (define-key dumb-jump-mode-map "l" #'dumb-jump-quick-look)
  (define-key dumb-jump-mode-map "b" #'dumb-jump-back))
```
So I can access all the dumb-jump commands after `C-c j` without enabling the mode.

My point is:

When are you planning to remove the legacy interface? If it is not soon, could you please add a custom to silent the warning under user knowledge?

Best,
Ergus

phikal:
> When are you planning to remove the legacy interface? If it is not soon, could you please add a custom to silent the warning under user knowledge?

I advocated for removing the legacy interface, but a lot of people have complained so I don't think we will be doing that any time soon. The only way I could see that happening, would be if we re-implemented the legacy interface, as a layer over the xref interface, but that would come at the cost of changing the UI.

#381 should implement something like what you suggested. It's simple (the option doesn't change anything, it just prevent obsoletion warnings from being installed), but it's only a first approximation.

phikal:
Did #381 fix the issue for you?

Ergus:
Hi:

Yes it partially solved the issue.

There are still some messages like

```
In toplevel form:
gnuplot-context.el:236:13: Warning: Package cl is deprecated

In gnuplot-match-pattern:
gnuplot-context.el:1820:30: Warning: ‘incf’ is an obsolete alias (as of 27.1);
    use ‘cl-incf’ instead.
[...]

Issue #378, opened 2020-11-15
dumb-jump-go still very useful


DavidVujic:
I very much like the new way of navigating with `M-.` and via `xref`. 
When in `clojure-mode` the `cider-find-var` command will be used, if I have understood it correctly.

I think the now deprecated `dumb-jump-go` (`C-M-g`) is still very useful:

* When browsing Clojure code without a nrepl/cider session running. 
* When writing ClojureScript and using `re-agent`, the `cider-find-var` doesn't jump to events or subscriptions (I guess because of them not being `Vars`). But `dumb-jump-go` does the job very good! 

Would it be a good idea to not deprecate that feature?

phikal:
Could you explain what `dumb-jump-go` is better at in the examples you gave, that the xref iterface isn't?

DavidVujic:
Here's an example - xref doesn't seem to find ClojureScript `re-agent` events or subscriptions. Here, I first use `C-.` with no findings and then `C-M-g` (dumb-jump-go) with a popup.

The behavior is slightly different when connected to a `Cider` session or not. However, dumb-jump-go will in both cases present a search result. In both cases, xref doesn't find anything. I have verified this on a Linux machine running Emacs 25, and on a Mac running emacs 26. Maybe there is a configuration I should add for xref?

![with-cider](https://user-images.githubusercontent.com/301286/99296702-72dff100-2847-11eb-9d1a-acc06df9ea94.gif)

phikal:
From what I see, xref doesn't seem to be using the dumb-jump backend (the miniprompt says "Visit tags table"), so it might be that `xref-backend-functions` has just changed. What does `M-: xref-backend-functions RET` say?

DavidVujic:
```
(defvar xref-backend-functions nil
  "Special hook to find the xref backend for the current context.
Each function on this hook is called in turn with no arguments,
and should return either nil to mean that it is not applicable,
or an xref backend, which is a value to be used to dispatch the
generic functions.")

;; We make the etags backend the default for now, until something
;; better comes along.  Use APPEND so that any `add-hook' calls made
;; before this package is loaded put new items before this one.
(add-hook 'xref-backend-functions #'etags--xref-backend t)
```

DavidVujic:
Oh! I realize I might have messed up my configuration. Accdentally put the 
```
;; jump to definition
(dumb-jump-mode)
(add-hook 'xref-backend-functions #'dumb-jump-xref-activate)
```
[...]

Issue #365, opened 2020-09-01
‘dumb-jump-go’ is an obsolete command (as of 2020-06-26)


avatar-lavventura:
I was using dumb-jump (https://github.com/jacktasia/dumb-jump) but suddenly it stops working.

I am getting following error:

```
‘dumb-jump-go’ is an obsolete command (as of 2020-06-26); 
‘dumb-jump-go’ has been obsoleted by the xref interface.

’function’ python  declaration not found.
```

-------

My setup:
```
(add-hook 'xref-backend-functions #'dumb-jump-xref-activate)
(global-set-key "\C-x\C-o" 'dumb-jump-go-prefer-external-other-window);
(global-set-key "\C-x\C-k" 'xref-pop-marker-stack);
(global-set-key "\C-x\C-j" 'dumb-jump-go);
(global-set-key "\C-x\C-k" 'dumb-jump-back);
```

jacktasia:
It should still be working even if it has been marked as obsolete. The real error here is `’function’ python  declaration not found.` which just means it could not find any results for whatever you were searching for.

```
def function():
    pass


function() # <-jumping from here worked for me.
```

avatar-lavventura:
@jacktasia it also prevents me to use ‘dumb-jump-back’ and automatically function converted into ‘xref-pop-marker-stack’.

```
‘dumb-jump-back’ is an obsolete command (as of 2020-06-26); 
‘dumb-jump-back’ has been obsoleted by ‘xref-pop-marker-stack’.
```

phikal:
What do you mean by "prevents"? Does it not jump back? Currently `dumb-jump-back` is not much more than a wrapper around `pop-tag-mark`, that in turn is just an alias for `xref-pop-marker-stack`, which is already bound to `M-.`. What happens when you press that key?

@jacktasia It might make sense to point that out in the deprecation string.

avatar-lavventura:
I can jump back using ‘dumb-jump-back’ that is aliased to ‘xref-pop-marker-stack’ and `pop-tag-mark` works too (as I understand).

But I cannot use ‘dumb-jump-go’ but when I try `elpy-goto-definition` , which works. So maybe if `elpy-mode` is on ‘dumb-jump-go’ could be aliased to `elpy-goto-definition`.

jacktasia:
[...]

Issue #353, opened 2020-07-18
[Question] How to combine with lsp-mode?


condy0919:
`lsp-mode` overwrite `xref-backend-functoins` using the following code:

``` elisp
(add-hook 'xref-backend-functions #'lsp--xref-backend nil t)
```

How to combine `dumb-jump` with `lsp-mode` so than when `lsp-mode` finds nothing <kbd>M-.</kbd> (`xref-find-definitions`) can fallback to `dumb-jump`?

phikal:
This doesn't seem that doable, as of now. Xref checks all functions in `xref-backend-functions` until the first one returns `nil`, but since LSP [doesn't do any checking](https://github.com/emacs-lsp/lsp-mode/blob/00fcee92ae2e57055fdb31d25741b9637fe96215/lsp-mode.el#L6174), it would always run.

The real problem is that xref doesn't (currently) have an option to say "if the first method fails, try the next function in `xref-backend-functions`".

But assuming that could be fixed (and it is doable, I think I remember this being mentioned on the emacs-devel mailing list), you'd probably just have to add a hook to lsp-mode, to add `dumb-jump-xref-activate` to `xref-backend-functions`, unless LSP does some additional magic.

amno1:
I was annoyed with all the "obsolete" compiler warnings I get and I am trying to understand why this deprecation. I also saw this thread, so I have a question:

Why deprecating dumb-jumps interface? Dumb-jump could just use xref as another "source" under the hood so to say. So methods like dumb-jumb-go could actually call xref-whatever which really is call into lsp behind the scene, and if it returns nada, it could do it's own internal magic and search gymnastics? We don't need lsp to call dumb-jumb as it's source. We will anyway bind this functionality to a key, so what function is bound to a key, like dumb-jumb-go or lsp-find-references or whatever it might be called is irrelevant.

phikal:
@amno1 The rationale was explained in #349 (and related/linked threads), so it's probably better to discuss the issue over there. Basically it boils down to not having to maintain multiple interfaces.

But to answer your immediate question, if you want silent Emacs' "obsolete" warnings, you can just unmark it as so

~~~el
(put 'dumb-jump-go 'byte-obsolete-info nil)
~~~

---

> Dumb-jump could just use xref as another "source" under the hood so to say.

[...]

PR #381, opened 2020-12-04
Allow silencing warnings about the legacy interface

phikal:
This patch would implement something as suggested in #380. It would probably make sense to add it, if we don't plan to remove the legacy interface.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -1705,6 +1705,11 @@ inaccurate jump).  If nil, jump without confirmation but print a warning."
   :group 'dumb-jump
   :type 'boolean)
 
+(defcustom dumb-jump-disable-obsolete-warnings nil
+  "If non-nil, don't warn about using the legacy interface."
+  :group 'dumb-jump
+  :type 'boolean)
+
 (defun dumb-jump-message-prin1 (str &rest args)
   "Helper function when debugging apply STR 'prin1-to-string' to all ARGS."
   (apply 'message str (-map 'prin1-to-string args)))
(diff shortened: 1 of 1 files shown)


dumb-jump.el
```diff
@@ -2980,24 +2985,25 @@ Using ag to search only the files found via git-grep literal symbol search."
...
-             dumb-jump-go-current-window))
-    (make-obsolete
-     obsolete
-     (format "`%s' has been obsoleted by the xref interface."
-             obsolete)
-     "2020-06-26"))
-  (make-obsolete 'dumb-jump-back
-                 "`dumb-jump-back' has been obsoleted by `xref-pop-marker-stack'."
-                 "2020-06-26")
+  (when dumb-jump-use-legacy-interface
```
jacktasia, line 2988: Should this be `unless` and not `when`?

dumb-jump.el
```diff
@@ -1705,6 +1705,11 @@ inaccurate jump).  If nil, jump without confirmation but print a warning."
   :group 'dumb-jump
   :type 'boolean)
 
+(defcustom dumb-jump-use-legacy-interface nil
```
jacktasia, line 1708: What do you think of `dumb-jump-disable-obsolete-warnings` instead of `dumb-jump-use-legacy-interface`? To be more specific about what this is actually doing since it's not actually required for using the legacy interface.

jacktasia (changes requested):
Thanks for opening this! I think this is a great change.

phikal:

> [1:text/plain Hide]
>
> @jacktasia requested changes on this pull request.
>
> Thanks for opening this! I think this is a great change.
>
>> -             dumb-jump-go-prefer-external-other-window
> -             dumb-jump-go-prompt
> -             dumb-jump-quick-look
> -             dumb-jump-go-other-window
> -             dumb-jump-go-current-window
> -             dumb-jump-go-prefer-external
> -             dumb-jump-go-current-window))
> -    (make-obsolete
> -     obsolete
> -     (format "`%s' has been obsoleted by the xref interface."
> -             obsolete)
> -     "2020-06-26"))
> -  (make-obsolete 'dumb-jump-back
> -                 "`dumb-jump-back' has been obsoleted by `xref-pop-marker-stack'."
> -                 "2020-06-26")
> +  (when dumb-jump-use-legacy-interface
>
> Should this be `unless` and not `when`?

Yes, you're right!

>> @@ -1705,6 +1705,11 @@ inaccurate jump).  If nil, jump without confirmation but print a warning."
>    :group 'dumb-jump
>    :type 'boolean)
>  
> +(defcustom dumb-jump-use-legacy-interface nil
>
> What do you think of `dumb-jump-disable-obsolete-warnings` instead of `dumb-jump-use-legacy-interface`? To be more specific about what this is actually doing since it's not actually required for using the legacy interface. 

Sounds good, I'll fix this too.

-- 
	Philip K.

jacktasia (approved):
lgtm. Thanks!

phikal:
Besides the already mentioned #380, this might also be interesting for #378, #365 and #353.

PR #389, opened 2021-02-21
Fennel lang support

andreyorst:
Hi! Thanks for amazing package!

This PR adds support for [Fennel language](https://fennel-lang.org/). It is a small source-to-source compiler for S-expression based syntax to Lua syntax. Because this compiles to Lua, and it can use Lua libraries as is, perhaps it should be able to jump to Lua code as well, but I was unable to find if dumb-jump can try other languages when nothing is found for current one.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -846,6 +846,21 @@ or most optimal searcher."
            :regex "\\bJJJ\(\\\(.+\\\)\)*\\s*="
            :tests ("test = osc + 0.5;" "test(freq) = osc(freq) + 0.5;"))
 
+    ;; fennel
+    (:type "variable" :supports ("ag" "grep" "rg" "git-grep") :language "fennel"
+           :regex "\\((local|var)\\s+JJJ\\j"
+           :tests ("(local test (foo)"
+                   "(var test (foo)"))
+
+    (:type "function" :supports ("ag" "grep" "rg" "git-grep") :language "fennel"
+           :regex "\\(fn\\s+JJJ\\j"
+           :tests ("(fn test [foo]")
+           :not ("(fn test? [foo]"))
+
+    (:type "function" :supports ("ag" "grep" "rg" "git-grep") :language "fennel"
+           :regex "\\(macro\\s+JJJ\\j"
+           :tests ("(macro test [foo]"))
+
     ;; fortran
     (:type "variable" :supports ("ag" "grep" "rg" "git-grep") :language "fortran"
            :regex "\\s*\\bJJJ\\s*=[^=\\n]+"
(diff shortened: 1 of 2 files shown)


dumb-jump.el
```diff
@@ -3,8 +3,8 @@
 ;; Author: jack angers and contributors
 ;; Url: https://github.com/jacktasia/dumb-jump
 ;; Version: 0.5.3
-;; Package-Requires: ((emacs "24.3") (s "1.11.0") (dash "2.9.0") (popup "0.5.3"))
-;; Keywords: programming
+;; Package-Requires: ((emacs "25.1") (s "1.11.0") (dash "2.9.0") (popup "0.5.3"))
```
phikal, line 6: You don't have to update this, the code that relies on Emacs 25 is only conditionally evaluated, if xref is available.

andreyorst: Ah, I see. Sorry! But what about `cl-defmethod`? Should I include cl-lib as requirement then?

phikal:
Andrey Listopadov <notifications@github.com> writes:

> Ah, I see. Sorry! But what about cl-defmethod? Should I include cl-lib as requirement then?

cl-defmethod is also only used in the same block, and cl-lib is already
required.

It is generally better to split issues like this up into multiple pull
requests, as this shouldn't have to do with Fennel.

-- 
	Philip K.

andreyorst:
> Andrey Listopadov <notifications@github.com> writes:
> Ah, I see. Sorry! But what about cl-defmethod? Should I include cl-lib as requirement then?
> cl-defmethod is also only used in the same block, and cl-lib is already required. It is generally better to split issues like this up into multiple pull requests, as this shouldn't have to do with Fennel.
> […](#)
> -- Philip K.

OK! :)

andreyorst:
should be ready, unless there's a way to add jumping to Lua sources as a fallback language search

phikal:
I might be mistaken, but that sounds like out-of-scope for dumb-jump, or can you just mix Lua and Fennel code in the same project?

andreyorst:
> can you just mix Lua and Fennel code in the same project?

Yes. TLDR Fennel == Lua => true

Fennel is simply a compiler from Fennel to Lua -- it outputs Lua code, so typically there's a mix of both languages, either because of embedding Fennel compiler into application that already uses Lua, or because of usage of some Lua library, which defines desired function or some variable. And given that Fennel compiler itself is a single Lua file, it may be nice to be able to jump to it's sources (although this may be rarely needed, but still). 

Here's an example of [small fennel project](https://gitlab.com/andreyorst/love-raymarching) that shares both fennel and Lua files. A more complex project may install libraries via Luarocks, a Lua package manager, directly into project tree, so again, it may be useful to jumpt to these directly. Basically saying in this regard Fennel == Lua. One thing to concern is that when doing AOT compilation of Fennel to Lua, thus producing essentially duplicate lua files (which duplicate definitions from fennel files) there may be confusion by default, but I think that this can be easily solved by adding .dumbjump file to the project with patterns to ignore.

But if there's no way to make this feel free to merge this as is, if everything is OK in the code.

jacktasia:
Thanks for opening this! This looks good to me. Let me check a bit more, but this should be cool to merge.

jacktasia (approved):
Thanks!

PR #391, opened 2021-02-27
Add initial support for Apex, the language for Salesforce servers

nxtr:
Apex is a strongly typed, object-oriented programming language, using
syntax that looks like Java and acts like database stored procedures.

https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_dev_guide.htm


jacktasia (approved):
Thanks!

Issue #135, opened 2017-05-16
Editing an org src block?


therockmandolinist:
It'd be nice if dumb-jump worked when using `org-edit-special` on an src block. Right now I get something like: `Could not find rules for '.org[ emacs-lisp ] *] file'.`

jacktasia:
Thanks for opening this. I don't currently use org-mode so this wouldn't be something I could easily add, but I'll happily take a PR if someone wants to add it.

therockmandolinist:
Ah, ok. Clarifying question: the README says dumb-jump can use the `major-mode` of the buffer - any reason why that wouldn't work here? In these cases the major mode is just the mode for the src language - so in the above example, `org-edit-special` just enters a buffer in `emacs-lisp-mode`.

jacktasia:
> Clarifying question: the README says dumb-jump can use the major-mode of the buffer - any reason why that wouldn't work here? In these cases the major mode is just the mode for the src language - so in the above example, org-edit-special just enters a buffer in emacs-lisp-mode

Sorry, the README isn't very specific, but basically, it uses major-mode only as a fallback, and it's honestly not that smart (because it's not used very often). That _may_ work in some cases with `org-edit-special`, but it won't work in the case of emacs lisp because dumb-jump classifies emacs-lisp as `elisp` and the major-mode for `emacs-lisp-mode` would come back as `emacs-lisp` (`elisp` != `emacs-lisp`)

Again, I don't use org-mode, but if you can provide a file and/or instructions for this `org-edit-special` scenario for emacs-lisp then I'll see if I can get this working...should be easy as including a map of `major-mode`s to the language name we're using (which for the most part already matchup except for this [elisp] and maybe a few other cases). Thanks!

NightMachinery:
In special edit buffers for Python, `(dumb-jump-get-language nil)` correctly returns `python`, but dumb-jump still does not work. I think this is because there is no actual file on disk for the backend (`dumb-jump-force-searcher`, `rg` in my case) to search? Is it possible to solve this?

PR #411, opened 2022-01-18
add org src block feature

Pierre-Andre:
This PR aims at providing a dumb-jump "definition search" in org-mode src-block. It closes (in some way) issue #135.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -251,6 +251,12 @@ or most optimal searcher."
   :group 'dumb-jump
   :type 'string)
 
+(defcustom dumb-jump-search-type-org-only-org
+  t
+  "If non nil restrict type ag/rg to org file.
+If nil add also the language type of current src block"
+  :group 'dumb-jump
+  :type 'boolean)
 
 (defcustom dumb-jump-find-rules
   '((:type "function" :supports ("ag" "grep" "rg" "git-grep") :language "elisp"
@@ -2038,9 +2044,9 @@ for user to select.  Filters PROJ path from files for display."
      ((and (eq dumb-jump-selector 'helm) (fboundp 'helm))
       (helm :sources
             (helm-make-source "Jump to: " 'helm-source-sync
-                                    :action '(("Jump to match" . dumb-jump-result-follow))
-                                    :candidates (-zip choices results)
-                                    :persistent-action 'dumb-jump-helm-persist-action)
+                              :action '(("Jump to match" . dumb-jump-result-follow))
+                              :candidates (-zip choices results)
+                              :persistent-action 'dumb-jump-helm-persist-action)
             :buffer "*helm dumb jump choices*"))
      (t
       (dumb-jump-to-selected results choices (popup-menu* choices))))))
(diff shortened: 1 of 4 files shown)


dumb-jump.el
```diff
@@ -2038,9 +2044,9 @@ for user to select.  Filters PROJ path from files for display."
      ((and (eq dumb-jump-selector 'helm) (fboundp 'helm))
       (helm :sources
             (helm-make-source "Jump to: " 'helm-source-sync
-                                    :action '(("Jump to match" . dumb-jump-result-follow))
-                                    :candidates (-zip choices results)
-                                    :persistent-action 'dumb-jump-helm-persist-action)
+                              :action '(("Jump to match" . dumb-jump-result-follow))
+                              :candidates (-zip choices results)
+                              :persistent-action 'dumb-jump-helm-persist-action)
```
phikal, line 2049: These seem to only be whitespace changes?

dumb-jump.el
```diff
@@ -2067,27 +2073,101 @@ to keep looking for another root."
...
+  "Return nil if NEWLANG  is already in PROPLIST or (if not)
+return a new proplist. The new proplis is PROPLIS
+where a NEWLANG plist(s) is (are) added to PROPLIST.
+The plist(s) value of NEWLANG is (are) copied from
+those of LANG and LANG is replaced by NEWLANG." 
+  (let ((alreadyuptodate
+         (--filter (string= newlang (plist-get it :language))
+                   proplist)))
+    (if alreadyuptodate
+        nil
```
phikal, line 2120: You can drop the let here and use an unless instead of if.

dumb-jump.el
```diff
@@ -2067,27 +2073,101 @@ to keep looking for another root."
...
+                                      (string= complang (plist-get it :language))
+                                      (string= extension (plist-get it :ext))
+                                      (string= agtype (plist-get it :agtype))
+                                      (string= rgtype (plist-get it :rgtype)))
+                                     dumb-jump-language-file-exts))
+         (newfindrule (dumb-jump-add-language-to-proplist complang dumb-jump-find-rules lang))
+         (newfileexts (dumb-jump-add-language-to-proplist complang dumb-jump-language-file-exts lang)))
+    ;; add (if needed) composite language to dumb-jump-find-rules
+    (if newfindrule
+        (set-default 'dumb-jump-find-rules newfindrule))
```
phikal, line 2142: A when is preferable to a single if used for side effects.

dumb-jump.el
```diff
@@ -2067,27 +2073,101 @@ to keep looking for another root."
...
+    ;; add (if needed) composite language to dumb-jump-language-file-exts
+    (if (not dumb-jump-search-type-org-only-org)
+        (if newfileexts
+            (set-default 'dumb-jump-language-file-exts newfileexts)))
```
phikal, line 2146: Same as above, just with unless.

dumb-jump.el
```diff
@@ -2067,27 +2073,101 @@ to keep looking for another root."
...
+    complang))
+
+(defun dumb-jump-get-language-from-aliases (lang)
+  "Extract the lang from aliases."
+  (let* ((lookup '(sh "shell" shell "shell" cperl "perl"
+                      matlab "matlab" octave "matlab"
+                      emacs-lisp "elisp" elisp "elisp"
+                      R "r" r "r"))
+         (result (plist-get lookup (intern lang))))
+    (if result result nil)))
```
phikal, line 2161: The if is redundant, but I'd also change the entire function to use assoc/alist-get or cond, instead of plist-get.

```
(alist-get lang '(("sh" . "shell") ("shell" . "shell") ...))
```

dumb-jump.el
```diff
@@ -2097,15 +2177,16 @@ to keep looking for another root."
...
       (plist-get (car result) :language))))
 
 (defun dumb-jump-issue-result (issue)
   "Return a result property list with the ISSUE set as :issue property symbol."
   `(:results nil :lang nil :symbol nil :ctx-type nil :file nil :root nil :issue ,(intern issue)))
 
 (defun dumb-jump-get-results (&optional prompt)
-  "Run dumb-jump-fetch-results if searcher installed, buffer is saved, and there's a symbol under point."
+  "Run dumb-jump-fetch-results if searcher installed, buffer is saved,
+and there's a symbol under point."
```
phikal, line 2189: The first line of a docstring should be a complete sentence (checkdoc).

phikal: I actually missed this in a few of the last docstring, that should be fixed too.

dumb-jump.el
```diff
@@ -2067,27 +2073,101 @@ to keep looking for another root."
...
+                          (--map (plist-get it :language)
+                                 dumb-jump-find-rules)))
         language
       (format ".%s file" (or (file-name-extension file) "")))))
 
+(defun dumb-jump-get-language-in-org ()
+  "In org mode, if inside a src block return
+associated language or org when outside a src block."
+  (let ((lang (nth 0 (org-babel-get-src-block-info))))
+                                        ; if lang exists then create a composite language
```
phikal, line 2101: Use a ;; comment here, the indentation if unnecessary.

phikal (changes requested):
Not much to say about the actual feature (I'm assuming it does what it should), so I'll just leave some style related comments.

phikal:
I don't think that fixing "unrelated" checkdoc issues should be part of this pull request.

Pierre-Andre:
I have undone all "unrelated checkdoc issues" on my local repos. I can push it if needed.

phikal:
That would be nice.

On 29 January 2022 15:27:31 CET, Pierre-Andre ***@***.***> wrote:
>I have undone all "unrelated checkdoc issues" on my local repos. I can push it if needed.
>

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

phikal (dismissed):

jacktasia:
Looks like the 24.x tests are failing with

`Symbol's value as variable is void: org-src--beg-marker`

phikal:
Using (org-)internal variables is bound to trigger issues eventually, but I don't know enough about Org internals to understand what they are being used for. Is there any alternative to using these variables directly? Having dumb-jump depend on a newer version of org would be a sad solution.

Pierre-Andre:
I will investigate on a better solution. I apologize  for this proposal that needs to be fine-tuned.

5 févr. 2022 22:55:20 Philip ***@***.***>:

jacktasia:
The 24.x tests hung. Retriggering them

Pierre-Andre:
Well I need to set up a docker image of emacs 24.5 and 24.4 to understand what's going on. It will take much more time…

Pierre-Andre:
emacs 24.x tests hung because
1. silversearch-ag version is too old (2.1.0 needed)
    -> fix: test are done with rg using `dumb-jump-force-searcher`
2. org-version is less than 9 and it implies that
   - there is a bug in `org-edit-src-exit` : when a src-edit buffer is openned with `org-edit-src-code` and no change is made and the src-buffer is closed with `org-edit-src-exit` the org buffer is seen (wrongly) as modified -> fix: I added (save-buffer) for these versions
   -  there another bug in dumb-jump determination of project-root and the test cannot be done properly -> fix: I set `dumb-jump-project`

I hope that all will be fixed but as docker tests with 24.5 does not work at all I cannot be sure…

Pierre-Andre:
Using the following Dockerfile
```
FROM debian
## basic downloads clients
RUN apt-get update && \
    apt-get --no-install-recommends install -y \
         ca-certificates \
         curl \
         git \
         openssh-client \
    && rm -rf /var/lib/apt/lists/*

## for emacs building process
RUN apt-get update && \
    apt-get --no-install-recommends install -y \
        autoconf \
        autogen \
        gcc \
        libgnutls28-dev \
        libncurses-dev \
        make \
        pkg-config \
        texinfo \
    && rm -rf /var/lib/apt/lists/*



######## silvergrep
RUN apt-get update && \
    apt-get --no-install-recommends install -y \
        silversearcher-ag  \
    && rm -rf /var/lib/apt/lists/*

######## ripgrep
RUN curl -LO https://github.com/BurntSushi/ripgrep/releases/download/0.10.0/ripgrep_0.10.0_amd64.deb && dpkg -i ripgrep_0.10.0_amd64.deb && rm ripgrep_0.10.0_amd64.deb && rm -f /var/cache/apt/archives/*.deb /var/cache/apt/archives/partial/*.deb /var/cache/apt/*.bin || true

######## Emacs
RUN cd / && mkdir -p /src && cd /src \
  && git clone --depth 1 git://git.sv.gnu.org/emacs.git emacs
 RUN cd /src/emacs   \
     && ./autogen.sh \
     && ./configure  --without-gif\
                          --with-x-toolkit=no \
                          --without-x \
                          --without-all \
                          --with-gnutls \
                          --prefix="/usr/local" \
     && make  && make install
# #
# ######### Cask
ENV PATH="/root/.cask/bin:$PATH"
RUN mkdir -p /CASK && cd /CASK \
     && git clone https://github.com/cask/cask /CASK/.cask
ENV PATH="/CASK/.cask/bin:$PATH"

######### misc
RUN mkdir -p DUMBJUMP
WORKDIR /DUMBJUMP


ENTRYPOINT ["bash"]
```
I can run the snapshot version and the 
- dumb-jump-goto-file-line-test failed (the `(mock (ring-insert * *))` throws an error and does not seem to be useful here)
- other tests passed 

(NB: EVM emacs-git-snapshot recipe does not work on my laptop.)

Conclusion: these check problems on "snapshot" are not coming from org-mode

NightMachinery:
Any updates?

Pierre-Andre:
Three years after, I still don't understand where is the problem for this PR. If some person in charge of this project can drop a comment to understand what is to be done... Or may be this PR is not welcome and therefore a simple sentence will be sufficient to close it.

In all the cases mentioned above, I would like to thanks all the members of this project on their work.

Pierre-Andre:
1. I have simplified the python-version from '3.9.21' to '3.9'. Even python-version '3' is ok but as I don't understand why using a specific version (such as '3.9.21') I have made the minimal change to make check working.
2. In line 2969 of `dumb-jump.el` file `(first (s-split` should be `(cl-first (s-split` 
3. If the previous line is OK then in `test/test-helper.el` file  the last two lines
    ```
    (eval-when-compile 
       (require 'cl))
    ```
     can be deleted 
Best regards

jacktasia (approved):
lgtm! Thanks!

jacktasia:
> > Boils down to if the tests pass and generally seems safe, then I am happy to merge, but I don't have the time or desire to help dig in and help with this feature. I explored adding this type of functionality many years ago, but it just required adding too much complexity, and I personally don't really use org-mode much anymore.
> 
> Using a "working github action" tests pass, thus my question: anything else to do ?

Sorry. I’ve been on vacation. I appreciate the work! Thanks again

Issue #399, opened 2021-06-17
Adding dumb-jump to NonGNU ELPA


phikal:
In a [recent discussion](https://mail.gnu.org/archive/html/emacs-devel/2021-06/msg00688.html) on emacs-devel, the topic of including new packages in NonGNU ELPA came up again. In case one is unfamiliar, it is the a repository that will be added to Emacs 28 by default, made for packages that do not need a copyright assignment.

I think that dumb-jump would be a great candidate to include, seeing as it is one of top packages on MELPA already. 

Would there be any objections to this?

phikal:
Oh, and this could also be seen as a chance to clean up some older code and dependencies.

jacktasia:
@phikal I am certainly open to this if there is a desire on their end. However, it seemed like dumb-jump having an external dependency (of the desired searcher) was potentially problematic in the past.

phikal:
Jack Angers ***@***.***> writes:

> @phikal I am certainly open to this if there is a desire on their
> end.

I suggested dumb-jump, among others, and was a positive response.

> However, it seemed like dumb-jump having an external dependency
> (of the desired searcher) was potentially problematic in the past.

How come? I mean, grep.el has a obvious external dependency, and a lot
of other packages depend on external tools.

-- 
	Philip K.

jacktasia:
I just had some vague memory, but I could certainly be wrong. Again, I am supportive if this is what people want.

phikal:
> Oh, and this could also be seen as a chance to clean up some older code and dependencies.

It would be possible to create a separate branch for the NonGNU ELPA variant and drop some legacy code + dependencies (mainly popup). Would you be interested in that?

PR #418, opened 2022-05-08
Drop the "s" dependency

phikal:
Hi,

this patch would remove s, replacing it with the [compat](https://elpa.gnu.org/packages/compat.html) package.  This change would allow for dumb-jump to be added to NonGNU ELPA, as mentioned in #399.

### DIFF
--- a/Cask
+++ b/Cask
@@ -7,4 +7,5 @@
  (depends-on "el-mock")
  (depends-on "noflet")
  (depends-on "ert-runner")
- (depends-on "undercover"))
+ (depends-on "undercover")
+ (depends-on "compat"))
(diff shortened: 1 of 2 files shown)


thisch:
> With this changeset I'd also like resign as collaborator, as I do not use dumb-jump personally any more.

what else do you use?

jacktasia:
> With this changeset I'd also like resign as collaborator, as I do not use dumb-jump personally any more.

Thanks for all of your help with everything @phikal!

jacktasia:
> I see that you have generated a lot of code that replicates functionality from s.el. Are you aware of the Compat library, that already provides backwards compatible definitions for functionality added in later releases of Emacs. Note that Compat requires at least Emacs 25.1, but I think that is reasonable considering that no major, supported distribution should be distributing a release of Emacs older than 26.1.


Thanks for the heads up. I will need to look into that more regarding bumping the floor of version support.

Issue #424, opened 2022-06-16
Missing rule: defmacro for elisp


chasecaleb:
Hi, I think a rule along the lines of this should be added:

```elisp
(add-to-list 'dumb-jump-find-rules
               '(:type "function" :supports ("ag" "grep" "rg" "git-grep") :language "elisp"
                 :regex "\\\(defmacro\\s+JJJ\\j"
                 :tests ("(defmacro test (blah)" "(defmacro test\n"
                 :not ("(defmacro test-asdf (blah)" "(defmacro test-blah\n" "(defmacro tester (blah)"
                       "(defmacro test? (blah)" "(defmacro test- (blah)")))
```

I've tested it out briefly and it seems to work, but it's certainly possible I'm overlooking something. @jacktasia what do you think?

jacktasia:
Looks good to me. Happy to merge a PR if you want to make one.

chasecaleb:
I was just looking through some of my old contributions and noticed this issue was still open. Since #425 was merged, this issue is resolved so I'm closing it :+1:

PR #425, opened 2022-06-18
Add elisp defmacro rule

chasecaleb:
Here's a PR for #424. I have to say, those tests are quite exhausti(ve|ng)!


jacktasia (approved):
Thanks!

PR #432, opened 2023-03-30
Fix xrefs for remote connections over TRAMP

### LINKED ISSUE #413: xref jump fails on Tramp opened files
I observe some strange behavior with ```xref``` backend which doesn't shows with obsolete ```dumb-jump-go```.
When working over Tramp, through ```xref```, jump to candidate goes in context of local machine, and opening local buffer as no such file exists.
The same jump to candidate through obsolete ```dumb-jump-go``` works just fine.
```xref-file-name-display``` has no power here too: project-relative, abs

gsingh93:
Haven't done much extensive testing with this, but it fixes #413 for me.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -2231,7 +2231,15 @@ of project configuration."
                                               cur-line-num parse-fn generate-fn)
                        search-paths))
 
-         (results (delete-dups (--map (plist-put it :target look-for) raw-results))))
+         (tramp-path-prefix (or (file-remote-p default-directory) ""))
+
+         (results (delete-dups
+                   (--map
+                    (progn
+                      (plist-put it :target look-for)
+                      (plist-put it :path
+                                 (concat tramp-path-prefix (plist-get it :path))))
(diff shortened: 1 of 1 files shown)


gsingh93:
@jacktasia would you be able to take a look at this to see if it's the proper fix?

jacktasia (approved):
Thanks! Sorry for taking forever!

prashantvithani:
This patch has actually broken `dumb-jump-go`. It blindly adds `tramp-path-prefix` to a relative path, which is already added [`dumb-jump--result-follow`](https://github.com/jacktasia/dumb-jump/blob/3c2ab8dfff3a10d3c5f2efd69cfbc81fb5dbbd39/dumb-jump.el#L2587-L2590) as part of `default-directory` (default directory is `ssh:host:path/to/dir` and relative path in the `results` is always corresponding to `default-directory`.

Reverting this patch locally fixes it.

jacktasia:
@coderabbitai please review

PR #435, opened 2023-05-16
Added some janet support

dleslie:
This adds some fairly basic support for the Janet programming language. Function definitions and variable definitions should work, but does not (yet) support more complex binding forms like let, loop, and seq.


jacktasia (approved):

jacktasia:
Thanks!

PR #439, opened 2023-07-10
Add generic rule for matching all clojure def types

### LINKED ISSUE #438: dumb-jump doesn't find all clojure definitions
Consider the function is listed in the following way - 

```clojure
(defn ^:metadata hello-world
  []
  (println "hello world"))
```  

and called in the following way - 
```clojure
(hello-world)
```

On invoking the `xref-find-definitions` function on `hello-world`. It doesn't list the above defined function.

The same happens when you try to find tags defined by `deftest`. Eg: - 

```clojure
(deftest hello-test
 ;;
)
```

npcoder2k14:
+ Fixes #438 
+ Existing rules didn't have support for `deftest` or a function with metadata.
+ Example - (defn ^:some-data say-hello [] (println "hello"))

Above function wouldn't be matched even if it is defined using `defn`.

+ Since we are only using one regex instead of multiple, it has become faster as well.

Test:
- Added new testcases

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,7 +27,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions/setup-python@v1.1.1
         with:
-          python-version: '3.7.17'
+          python-version: '3.7'
           architecture: 'x64'
       - uses: purcell/setup-emacs@master
         with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,7 +27,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions/setup-python@v1.1.1
         with:
-          python-version: '3.7.16'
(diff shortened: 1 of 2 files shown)


npcoder2k14:
@jacktasia Could you please review this PR?

npcoder2k14:
Thanks for the response @jacktasia. I'll fix it in a couple of days.

npcoder2k14:
@jacktasia I have fixed the commit and ran all tests locally. 🤞 
Please check it now

npcoder2k14:
Builds are failing due to the python version upgrade.

jacktasia:
> Builds are failing due to the python version upgrade.

Interesting, because the initial run didn’t work because the old version wasn’t there any more. Now the new version it wanted before seems to be gone.

jacktasia (approved):
Thanks!

PR #445, opened 2024-01-14
support solidity contract

Chipe1:
Rule based on https://docs.soliditylang.org/en/latest/grammar.html#a4.SolidityParser.contractDefinition


jacktasia (approved):
Thanks!

PR #460, opened 2025-02-21
Add odin support

Lain62:
Added a simple odin lang support https://odin-lang.org/

it seems to be working well.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -1532,21 +1532,20 @@ or most optimal searcher."
 
     ;; odin
     (:type "function" :supports ("ag" "grep" "rg" "git-grep") :language "odin"
-           :regex "\\s*\\bJJJ\\s*::\\s*proc"
+           :regex "\\s*\\bJJJ\\s*:\\s*:\\s*proc"
            :tesots ("test :: proc()"
                    "test ::proc() {"
                    "test:: proc(a: i32) -> i32 {"
-                   "test::proc{}"))
-    
-    (:type "type" :supports ("ag" "grep" "rg" "git-grep") :language "odin"
-           :regex "\\s*\\bJJJ\\s*::\\s*(struct|enum|union)"
-           :tests ("test :: struct"
-                   "test ::enum"
-                   "test:: union"))
+                   "test::proc{}"
+                   "test: :proc "contextless"{}"))
     
     (:type "variable" :supports ("ag" "grep" "rg" "git-grep") :language "odin"
-           :regex "\\s*\\bJJJ\\s*:(\\s*[^=\\n]+\\s*:|:|\\s*[^=\\n]+\\s*=|=)"
-           :tests ("test :: [2]f32"
+           :regex "\\s*\\bJJJ\\s*:"
+           :tests ("test :: struct"
+                   "test ::enum"
+                   "test:: union"
+                   "test: : custom_type"
+                   "test :: [2]f32"
                    "test : f32 : 20"
                    "test: i32 : 10"
                    "test := 20"
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -1532,7 +1532,7 @@ or most optimal searcher."
 
     ;; odin
     (:type "function" :supports ("ag" "grep" "rg" "git-grep") :language "odin"
-           :regex "\\s*\\bJJJ\\s*:\\s*:\\s*proc"
+           :regex "\\s*\\bJJJ\\s*:\\s*(:|=)\\s*proc"
            :tesots ("test :: proc()"
                    "test ::proc() {"
                    "test:: proc(a: i32) -> i32 {"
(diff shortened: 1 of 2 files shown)


dumb-jump.el
```diff
@@ -1528,7 +1528,32 @@ or most optimal searcher."
     (:type "type" :supports ("ag" "grep" "rg" "git-grep") :language "apex"
            :regex "(class|interface)\\s*JJJ\\b"
            :tests ("class test:" "public class test implements Something")
-           :not ("class testnot:" "public class testnot implements Something")))
+           :not ("class testnot:" "public class testnot implements Something"))
+
+    ;; odin
+    (:type "function" :supports ("ag" "grep" "rg" "git-grep") :language "odin"
+           :regex "\\s*\\bJJJ\\s*::\\s*proc"
```
diniamo, line 1535: There may be whitespaces between the 2 colons as well.

Odin procedure declarations may specify an explicit calling convention. See here for more information: https://odin-lang.org/docs/overview/#calling-conventions

dumb-jump.el
```diff
@@ -1528,7 +1528,32 @@ or most optimal searcher."
...
+           :tesots ("test :: proc()"
+                   "test ::proc() {"
+                   "test:: proc(a: i32) -> i32 {"
+                   "test::proc{}"))
+    
+    (:type "type" :supports ("ag" "grep" "rg" "git-grep") :language "odin"
+           :regex "\\s*\\bJJJ\\s*::\\s*(struct|enum|union)"
```
diniamo, line 1542: Keep in mind that type aliases exist, so this won't work for cases like

```odin
Random_Name :: My_Type
```

Sadly, that also means that there is no precise regex we could use here.

diniamo (changes requested):

dumb-jump.el
Lain62: since a precise regex couldn't be used. maybe its better if i just try to generalize the regex more so it can capture either types or variables
something like `\\s*\\bJJJ\\s*:` since it seems all odin declarations either it be type, const, or variable (even proc) start with `name :` 
ill keep the functions seperate for now

dumb-jump.el
diniamo: Oh you're right

dumb-jump.el
diniamo: If anything, the types should be kept separate, and not the functions.

```odin
t := int
fmt.println(t)
t = uint
fmt.println(t)
```

is valid, while

```odin
t := int
fmt.println(t)
t = uint
fmt.println(t)
```
isn't, because types can only be compile-time constants.

diniamo:
@Lain62 what will it be then? `\\s*\\bJJJ\\s*:` should be used if we want to be 100% correct. However a trade-off could be made, because realistically you wouldn't search for functions and types that aren't declared in the `name :: (proc)` format, meaning we could go back to something like what you did originally, greatly reducing the number of results. Not sure if the second option is allowed in this project, but if it is, and you decide to do that, make sure to add a comment in the code, explaining this.

Lain62:
I think its more important to be 100% correct, and to have the biggest coverage. Don't really want a situation where people get frustrated for it not finding something even if its rare. So I think keeping it as it is, is the best.

diniamo:
But it's not 100% right now, since functions can also be declared with `=`.

diniamo:
Oh and on more thing, I would change the generic regex to include the optional type specifier and the second `:`/`=`. Might avoid some more cases.

jacktasia:
Thanks @Lain62 and @diniamo! All the tests pass, and this looks like a safe change, so I am ready to merge whenever you both are happy, assuming those two things stay true.  So, please let me know; also, If you want to add "Odin" to the `README.md` language list, that would be cool, or I can do it post-merge. Thanks again

Lain62:
alright I've added `=` to functions (forgot this exists) and added the optional type specifier and the second `:`/`=` for the generic regex. let me know if there's anything that needs changing.
also added odin to readme

diniamo:
Functions can also have an identifier, e.g., `var: proc() : proc() {}` and `var: proc() = proc() {}` are valid. I'm not sure I see the point of separating functions and variables.

Lain62:
okay ngl, i did not know you can put identifiers on functions.
removed it now

diniamo (approved):
Alright, I think that should work. I don't exactly understand the part of the regex that matches the type identifier, but it seems to have resolved every obscure test case I could throw at it, so we should be good.

I will make a note of this, but I'm fairly certain that we won't be able to handle very obscure cases like this:

```odin
var : struct {
    a: i32
} : { a = 1 } fmt.println(var)
```

is valid Odin. However as I said, I don't think we can account for that.

Lain62:
yeah i don't think its possible to account for it. i think it should be good now

diniamo:
@jacktasia Looks like we are ready to go.

jacktasia (approved):
Thanks!

PR #464, opened 2025-08-22
Add jai support

paylanon:
Hey :) thanks for the S-tier package.

Language is still in closed beta, but we've got a few spartan emacsers who could use the support.


jacktasia (approved):
lgtm, thanks!

Issue #448, opened 2024-04-12
`git grep` is preferred over `rg` but is massively deficient


zenspider:
Without any config, if any project I use that is backed by git, then dumb-jump prefers to use `git grep` to find definitions via `dumb-jump-pick-grep-variant`. The problem, in C or ruby or anything else I've tried, is that it doesn't find ANYTHING... 0% hit rate. If I set `'(dumb-jump-force-searcher 'rg)` then everything works well. I don't know if `git grep` regexps are deficient or broken because I find dumb-jump hard to navigate but the problem certainly lies with the searcher. For some reason I can't figure out, it chooses `git-grep` over `dumb-jump-prefer-searcher`, so I'm not sure what the `prefer` vs `force` distinction is for.

I would also suggest that rg be preferred over ag.

gmichokostas:
I've got the same issue as well. If I add `(setq dumb-jump-force-searcher 'rg)` in my config everything works.

mrunhap:
Same here, it just don't work at all unless set force searcher, tested on python and golang.

dpassen:
Duplicate of #428

zenspider:
I don't consider this a duplicate. Part of the problem, not addressed in the other issue, is ignoring my preference and choosing git grep over my settings.

zenspider:
Since this project doesn't seem to be maintained these days, I'm experimenting with a fork I'm making called `dumber-jump`. Only uses rg and completing-read and removes the long obsoleted functions. Much less to configure.

jacktasia:
> I've got the same issue as well. If I add `(setq dumb-jump-force-searcher 'rg)` in my config everything works.

This is the solution if git grep doesn’t work for you. I am really confused why this one line of config is such a big deal. 

The reason preferred (more like fallback) doesn’t work is because git grep is actually available and at some point git grep was considered the optimal searcher, but unfortunately git has borked its grep implementation in reason versions. It’s complicated by the fact that there are/were a very wide assortment of git versions in the wild, so I am hesitant to change that default. 

This should actually default to `grep` since that’s widely installed in a “just works” sense. 

No great solution in my mind. If someone wants to make a fork then that’s certainly fine with me.

pierre-rouleau:
@zenspider , I must admit I don't understand the problem either.  The code has the ``dumb-jump-force-searcher`` user-option.  It can be customized and takes over. What does "it ignores my preference" mean?

zenspider:
[...]

PR #482, opened 2026-01-12
Fix errors, byte-compiler warnings, docstring checkdoc reported warnings - parts 1, 2, 3, 4, 5.

pierre-rouleau:
* First part of docstring reformatting.
  - Includes an update to `dumb-jump-prefer-searcher` docstring that refers to its dumb-jump-force-searcher` counterpart (which I'm planning to extend as described at the end of https://github.com/jacktasia/dumb-jump/issues/448 in: https://github.com/jacktasia/dumb-jump/issues/448#issuecomment-3716529700) The change for that will come in a future commit.

The changes are mostly simple changes to reduce the line lengths but it gives a taste of what I'm planning to change in the remaining of the file.  Let me know if there are problems here and I'll adjust.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -2295,7 +2295,7 @@ This is the persistent action (\\[helm-execute-persistent-action]) for helm."
           (s-trim (plist-get result :context))))
 
 (defun dumb-jump-ivy-jump-to-selected (results choices _proj)
-    "Offer CHOICES as candidates through `ivy-read'.
+  "Offer CHOICES as candidates through `ivy-read'.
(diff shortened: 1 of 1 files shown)


pierre-rouleau:
Note that in fix part 4, I also removed the **second** string in `dumb-jump-process-results` and merged it inside the first one making a single docstring.  I forgot to mention it in the commit text

pierre-rouleau:
Hi Jack (@jacktasia), This last commit completes the first set of changes.  Once this is integrated I will proceed with submission of the code change described here: https://github.com/jacktasia/dumb-jump/issues/448#issuecomment-3716529700

PR #488, opened 2026-01-15
Change searcher user-option behaviour: force is now an override

pierre-rouleau:
* The main purpose of this commit is an attempt to address [issue 448](https://github.com/jacktasia/dumb-jump/issues/448) while keeping backward compatibility to the current value of the `dumb-jump-prefer-searcher` and `dumb-jump-force-searcher` user-options.

  The idea is to turn the `dumb-jump-force-searcher` user-option into an override of `dumb-jump-prefer-searcher` under some situations to allow the user to select the searcher tool preferred for all the project *except* the ones that meet the conditions identified by the value or the `dumb-jump-force-searcher` user option.

  To do that the code adds 2 new `dumb-jump-force-searcher` choices:

  - A user-specified *function*.  That function must accept one argument, the name of the project directory. The function returns a symbol that identifies the required searcher tool (something like `'grep` or `'ag`) if one is specified or nil if the function determines that the searcher selection for this project must not be overrident (and selected by the `dumb-jump-prefer-searcher` value).

  - A list of project directories (strings) where dumb-jump must use git grep as the searcher tool.

  A nil value for `dumb-jump-prefer-searcher` now means that dumb-jump will
  honour the searcher tool selection made by the value of
  `dumb-jump-prefer-searcher`.

  The above 3 choices should be the only ones allowed.  However, to continue
  supporting the behaviour dumb-jump users have come to learn (always use
  `dumb-jump-force-searcher` to select the search tool they want because
  previously dumb-jump would prioritize using git grep) the a transition
  period must keep the support of the other original values (the tool name
  symbols). The tags of those values are identifying them as "deprecated" and
  users are discouraged to use them since they are no longer needed.

* To help testing the code splits the `dumb-jump-pick-grep-variant` function into 2 functions: the top one remains `dumb-jump-pick-grep-variant` but it calls `dumb-jump-selected-grep-variant` which now holds the complete searcher tool selection logic based on the optionally passed project directory and the values of the 2 user-options and returns the searcher tool symbol.

* Also wrapped long line in `xref-backend-identifier-completion-table`.

* Inside test/dumb-jump-test.el a change to the corresponding ERT tests code supports the new searcher tool selection logic, testing all combinations.

  - Also formatted long lines of some tests.
  - Also fixed the only remaining byte-compiler warning in `dumb-jump-go-too-long-test` complaining about the unused prompt argument by renaming it to `_prompt`.

  - The old `dumb-jump-pick-grep-variant-git-grep-in-git-repo` test is no longer needed.  For the moment it is just commented-out and a comment
    before it describes why it is no longer needed.   A future commit could
    remove those comment lines.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -3,7 +3,7 @@
 ;; Author: jack angers and contributors
 ;; Url: https://github.com/jacktasia/dumb-jump
 ;; Version: 0.5.4
-;; Package-Requires: ((emacs "24.3") (s "1.11.0") (dash "2.9.0") (popup "0.5.3"))
+;; Package-Requires: ((emacs "24.4") (s "1.11.0") (dash "2.9.0") (popup "0.5.3"))
 ;; Keywords: programming
 
 ;; Dumb Jump is free software; you can redistribute it and/or modify it
@@ -136,15 +136,10 @@ one of two preferred choice overriding methods:
 - Override preferred for directories needing git-grep:
    Specify one or more project directories where the git-grep search tool must
    be used.
-
-Note:
-This currently continue to support the now deprecated original choices
-of tools: \\='ag, \\='rg, \\='grep, \\='gnu-grep, \\='git-grep, or
-\\='git-grep-plus-ag.
-Please do not use those, as they will no longer be supported in a future
-version of dumb-jump.  Select these in `dumb-jump-prefer-searcher' instead and
-set `dumb-jump-force-searcher' to nil to honour `dumb-jump-prefer-searcher'
-choice."
+- Override to a specific tool.  That is mostly useful to set the overriding
+  inside the .dir-locals.el of the directory.
+  The choices are: \\='ag, \\='rg, \\='grep, \\='gnu-grep, \\='git-grep, or
+                   \\='git-grep-plus-ag."
   :group 'dumb-jump
   :type '(choice (const
                   :tag "nil: honour `dumb-jump-prefer-searcher' choice." nil)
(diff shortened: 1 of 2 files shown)


pierre-rouleau:
@jacktasia 
Was there a `force` before?  
Currently in my implementation there is no `force`.  However you can still set `ag`, `rg` or the other to force using those as you were able to do before.  And that's what is considered "*deprecated*" but still available.

dumb-jump.el
```diff
@@ -116,22 +117,48 @@ unless that is nil."
...
                  (const :tag "git grep" git-grep)
                  (const :tag "git grep + ag" git-grep-plus-ag)))
 
-(defcustom dumb-jump-force-searcher
-  nil
-  "Forcibly use searcher: \\='ag, \\='rg, \\='git-grep, \\='gnu-grep, or \\='grep.
-Set to nil to not force anything and use `dumb-jump-prefer-searcher'
-or most optimal searcher."
+(defcustom dumb-jump-force-searcher nil
+  "Search tool override of `dumb-jump-force-searcher' selection.
```
jacktasia: shouldn't this be "override of `dumb-jump-prefer-searcher`"?

pierre-rouleau:
@jacktasia  I just saw your notes about the Readme.  I think what you wrote would be fine.  
And if you want to never deprecate the forcing of a specific tool, it would work too.  
So I'm fine with what you wrote.

Do you want me to integrate it in this PR?

jacktasia:
> Do you want me to integrate it in this PR?

If you don't mind that'd be great so they all land together, but if you don't have the time totally fine and I can do it later.

pierre-rouleau:
@jacktasia In the end I was able to do it tonight.

jacktasia (approved):
Thanks!

Issue #498, opened 2026-01-25
[Question/Idea] Should dumb-jump check if ripgrep supports PCRE2?


pierre-rouleau:
Ripgrep command `rg` can be built with or without PCRE2 support.  Dumb-jump requires the ability to to use PCRE2, otherwise the ripgrep command will fail.  This has already been experienced and reported before (see https://github.com/jacktasia/dumb-jump/issues/128).  

The current dumb-jump code assumes that ripgrep supports PCRE2 but does not check for that.  The function `dumb-jump--rg-installed?` checks for ripgrep version but could also check if PCRE2 support is included.   If not it could report that rg is not present and perhaps report the reason why, in a message perhaps, but maybe also, inside a variable that the user could consult later (via a command function?)

PR #500, opened 2026-01-26
dumb-jump-rg-installed? now return nil when Ripgrep does not support PCRE2

pierre-rouleau:
This implements the suggestion in [issue #498](https://github.com/jacktasia/dumb-jump/issues/498).

* `dumb-jump-rg-installed?` now checks that RipGrep `rg` supports PCRE2.  The function returns nil (and memoize that) if it does not.

  - Any problem detected while checking for Ripgrep are stored inside the `dumb-jump--detected-env-problems` list.
  - Use the new command `dumb-jump-show-detected-problems` to print and erase the list of detected problem in the `*Message*` buffer.  If the command is issued with a prefix argument it does not erase the list.
  - Update ert tests accordingly

Note:  I ran tests locally and they all passed.  The Github workflow controlled tests take forever, a problem in the tool chain probably.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     timeout-minutes: 30
     strategy:
       fail-fast: false


jacktasia:
> @jacktasia Is there anything I should do to make these tests proceed?

Hmm, I thought some GitHub Actions (transitive?) dependency had messed things up because when I originally opened a dummy PR, it also hung, and it seemed fine once I modernized the `test.yml` 

but after pulling that change in it's still hanging and seemingly always on `dumb-jump.el: Warning: Use -zip-pair instead of -zip to get a list of pairs`

maybe we make that change here too? https://github.com/jacktasia/dumb-jump/blob/ce40eafb4c15f4c402783d29a38f54efbf6bec1e/dumb-jump.el#L3010

PR #502, opened 2026-01-28
dumb-jump-rg-installed? return nil if Ripgrep PCRE2 is not available + Fix -zip and stalling tests.

pierre-rouleau:
- Complete https://github.com/jacktasia/dumb-jump/pull/500
- Include all *.el code changes from https://github.com/jacktasia/dumb-jump/pull/500 to see if the tests will now pass as they did for the commit of the previous commit (however they again seem to be stalling, but not on the byte compilation: cask exec ert-runner seems to stall...)

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -2730,25 +2730,25 @@ variable `dumb-jump--detected-env-problems'."
   (if (eq dumb-jump--rg-installed? 'unset)
       (setq
        dumb-jump--rg-installed?
-       (catch 'test-fail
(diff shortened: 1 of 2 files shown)


pierre-rouleau:
OK, I got the non-disabled tests to pass by adding mocking of `dumb-jump-rg-installed?` to a lot of tests.
Now I'll start restoring the tests that test it and its immediate caller.

jacktasia (approved):
Nice! Thanks!

Issue #504, opened 2026-01-29
The validity of mocking searcher tool presence depends on the actual presence of the tool in the tested environment.


pierre-rouleau:
The Ert testing of the dumb-jump code done by test/ dumb-jump-test.el test several function that may or may not call the search tool presence checker function like `dumb-jump-rg-installed?`.  In some cases they are called when another tool is or is not present simply because of the logic inside dumb-jump.el

The test/ dumb-jump-test.el uses el-mock for mocking.   When a function is mocked by a (mock) form the mocked function **must** be called by the test code, otherwise it's reported as a test failure.
- This is where the problem lies:  a function like `dumb-jump-rg-installed?` may or may not be called depending on the presence of the other search tools in the test environment.  And test environment will differ depending on what search tool (or version of the search tool) is installed.

Fortunately, the el-mock library provides a solution to this problem:  the `(stub)` form is similar to the `(mock)` form but it is not considered a test failure when a stubbed function is not called.

The test environment used in the GitHub workflow happens to differ than mine in this level.  I got several errors after I fixed the test stalling issues that occurred in GitHub.    But the fixes was to introduce severl `(mock)` forms that cause test failures in my test environment.  They must all be replaced by stub forms.

The tests that fails in my environment are:
```
   FAILED  dumb-jump-a-back-test
   FAILED  dumb-jump-cpp-issue87
   FAILED  dumb-jump-cpp-test1
   FAILED  dumb-jump-cpp-test2
   FAILED  dumb-jump-fetch-results-test
   FAILED  dumb-jump-go-clojure-asterisk-test
   FAILED  dumb-jump-go-clojure-no-asterisk-test
   FAILED  dumb-jump-go-clojure-no-question-mark-test
   FAILED  dumb-jump-go-clojure-question-mark-test
   FAILED  dumb-jump-go-current-window-test
   FAILED  dumb-jump-go-include-lib-test
   FAILED  dumb-jump-go-js-es6-class-test
   FAILED  dumb-jump-go-js-es6a-test
   FAILED  dumb-jump-go-js-es6b-test
   FAILED  dumb-jump-go-js-es6c-test
   FAILED  dumb-jump-go-js-es6d-test
   FAILED  dumb-jump-go-js-es6e-test
   FAILED  dumb-jump-go-js2-test
   FAILED  dumb-jump-go-no-result-test
   FAILED  dumb-jump-go-no-rules-test
   FAILED  dumb-jump-go-nosymbol-test
[...]

PR #505, opened 2026-01-29
Fix failing test with mock replaced by stub

pierre-rouleau:
* This allows test passing when a function may or may not be called.
* See [issue 504](https://github.com/jacktasia/dumb-jump/issues/504)


jacktasia (approved):
Thanks!

Issue #499, opened 2026-01-25
[Suggestion] Add the ability to specify several comment strings for a language.


pierre-rouleau:
Currently the `dumb-jump-language-comments` user-option only supports defining *one* comment string for a specific language.
- There is nothing in the current dumb-jump code that prevents a user to add multiple entries inside the user-option but the code only uses the first one and silently ignores the others.
- It would be useful to be able to define more than one since the code just looks if the stripped found line starts with the comment string and excludes it.   We could, for example, specify both "//" and "/*" for the "c++" language. 
- Also note that allowing multiple comment string would allow the user to add any other contextual string that might help reduce the potential results under special circumstances.

Here's an implementation I am using:

``` lisp
(defun dumb-jump-get-comment-by-language (lang)
  "Return a list of the one-line comment or comments for the given LANG."
  (let* ((entries (-distinct
                   (--filter (string= (plist-get it :language) lang)
                             dumb-jump-language-comments))))
    (if (> (length entries) 0)
        (mapcar (lambda (pl) (plist-get pl :comment)) entries)
      nil)))

(defun dumb-jump--context-startswith-comment (context-string comments)
  "Return t when CONTEXT-STRING starts with any COMMENTS string."
  (let ((startswith-comment nil)
        ;; trim leading white-space from context-string                                                                                      
        (context-str (replace-regexp-in-string "^\\s-+" "" context-string)))
    (catch 'found-one
      (dolist (comment comments)
        (when (string-prefix-p comment context-str)
          (setq startswith-comment t)
          (throw 'found-one nil))))
    startswith-comment))

(defun dumb-jump-filter-no-start-comments (results lang)
  "Filter out RESULTS with a :context starting with a LANG comment in file."
  (let ((comments (dumb-jump-get-comment-by-language lang)))
    (if comments
        (-concat
         (--filter (not (dumb-jump--context-startswith-comment
                         (plist-get it :context)
                         comments))
                   results))
      results)))
```

The changes needed to implement this logic is relatively simple: no change to the `defcustom dumb-jump-language-comments` is required (except perhaps updating its docstring), just modify the `dumb-jump-get-comment-by-language`  and its only caller `dumb-jump-filter-no-start-comments`.

- `dumb-jump-language-comments` can now hold several comment entries for a language.  For example, the "c++" language can now support both "//" and "/*" as comment starts.
- The `dumb-jump-get-comment-by-language` function now returns a list of comment strings.
[...]

PR #506, opened 2026-01-29
Add ability to specify several comment strings for a language.

pierre-rouleau:
This addresses [issue 499](https://github.com/jacktasia/dumb-jump/issues/499)

* It is now possible to enter more than one comment string for a given language in the `dumb-jump-language-comments` user-option.
  - This is useful in languages like C++ where both `//` and `/*` may begin a comment line.  The second one was added to the default list value.
  - In some cases the user might want to temporary add extra strings in there to exclude some entities to be included as potential targets.
  - For the implementation of this feature: modified `dumb-jump-get-comment-by-language` and `dumb-jump-filter-no-start-comments` which now calls the new function `dumb-jump--context-startswith-comment`.

Also updated the following:

* Referenced `dumb-jump-debug` in `dumb-jump-quiet` docstring (and vice-versa).
* Describe `.dumbjumpignore` file in `dumb-jump-project-denoters` docstring.
* Other minor docstring enhancement.

### DIFF
--- a/test/dumb-jump-test.el
+++ b/test/dumb-jump-test.el
@@ -333,7 +333,7 @@
 
 (ert-deftest dumb-jump-run-cmd-test ()
   (with-mock
-    (mock (dumb-jump-rg-installed?) => t)
+    (stub dumb-jump-rg-installed? => t)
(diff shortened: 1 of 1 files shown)


jacktasia (approved):
Thanks!

PR #514, opened 2026-02-05
Improve regexp test

pierre-rouleau:
Hi @jacktasia , here's a collection of changes to the test code and dumb-jump.el to reduce duplication of regular expression strings in the test code while also making some minor improvements to the dumb-jump.el code.

The changes are better described in the commits.  The CodeRabbit summary is missing the essence so I will write it here (bizarrely I can't update the PR title, I thought I was able to do that before...).

* Eliminate duplication of regex strings inside the test code file by placing them inside one defconst instance and writing a function that provides the appropriately filled list to the test functions.
* Use the `--` option/argument separator in the searcher command line, eliminating the need to check if the regex starts with an hyphen.
* In the elisp rules of `dumb-jump-find-rules`, eliminate the 3rd backslash. It is not necessary and Emacs properly renders it with the font-lock-warning-face.  I have not yet done the same change for other languages.  I did it for elisp because I have tested dumb-jump operation manually to confirm that it continues to work properly.
* Added support for several function, methods, macro, structure definition forms in elisp: cl-defstruct, etc..
* Lined up some of the multiple-lines multiple-column forms to make them look like tables to help clarity.

## Summary by CodeRabbit

* **New Features**
  * Option to control how whitespace is treated in search regexes for improved cross-platform matching

* **Bug Fixes**
  * More robust handling of empty or special regex cases across search backends
  * Improved word-boundary handling to reduce false positives

* **Improvements**
  * Broader language pattern coverage for Elisp-based navigation
  * Updated behavior and messaging for search command generation

* **Tests & Docs**
  * Consolidated test templates and updated expectations; expanded documentation and inline notes

### DIFF
--- a/test/dumb-jump-test.el
+++ b/test/dumb-jump-test.el
@@ -1897,7 +1897,7 @@ VARIANT must be one of: ag, rg, grep, gnu-grep, git-grep, or git-grep-plus-ag."
       (forward-line 4)
       (forward-char 2)
       (with-mock
-        (stub dumb-jump-ag-installed? => t)
+        ;; (stub dumb-jump-ag-installed? => t)
         (stub dumb-jump-rg-installed? => t)
         (mock (dumb-jump-goto-file-line * 4 9))
         (should (string= clj-to-file (with-no-warnings (dumb-jump-go))))))))


jacktasia (commented):
Thanks!

Issue #513, opened 2026-02-05
Another case of hanging tests: dumb-jump prompts the user!


pierre-rouleau:
While testing regular expressions and modifying them I ran into a test hanging.  In my case it was `dumb-jump-go-clojure-no-asterisk-test` but really that does not matter much.  I issued the following command to run the test from the command line:

``` shell
cask exec ert-runner -p dumb-jump-go-clojure-no-asterisk-test
```

I also instrumented the test to activate debug tracing, this way:

``` elisp
(ert-deftest dumb-jump-go-clojure-no-asterisk-test ()
  (let ((clj-jump-file (f-join test-data-dir-proj3 "file3.clj"))
        (clj-to-file (f-join test-data-dir-proj3 "file2.clj"))
        (dumb-jump-debug t))            ;;;  <<<<========= TEMPORARY DEBUG TRACING                                                             
    (with-current-buffer (find-file-noselect clj-jump-file t)
      (funcall 'lisp-mode) ;; need a built in lisp mode so `?` is part of a symbol                                                     
      (goto-char (point-min))
      (forward-line 4)
      (forward-char 2)
      (with-mock
        (stub dumb-jump-ag-installed? => t)
        (stub dumb-jump-rg-installed? => t)
        (mock (dumb-jump-goto-file-line * 4 9))
        (should (string= clj-to-file (with-no-warnings (dumb-jump-go))))))))
```

And also added issuing a message just before the `dumb-jump-prompt-user-for-choice` call inside `dumb-jump-handle-results`:

``` elisp
    (dumb-jump-debug-message
     look-for
     ctx-type
     var-to-jump
     (pp-to-string match-cur-file-front)
     (pp-to-string results)
     prefer-external
     proj-root
     cur-file)
    (cond
     (use-tooltip ;; quick-look mode                                                                                                                                                                                                                                                       
      (popup-menu* (--map (dumb-jump--format-result proj-root it) results)))
     (do-var-jump
      (dumb-jump-result-follow var-to-jump use-tooltip proj-root))
     (t
      (progn
        (message "UNDER TEST: about to prompt the user!!!").  ;;; <<<<============ HERE
        (dumb-jump-prompt-user-for-choice proj-root match-cur-file-front))))))
```

The result from the test shows that several candidates are found and the code ends up trying to prompt the user, which hangs the test:
``` shell
[...]

PR #516, opened 2026-02-08
Fix issue #513: prevent prompting from hanging tests. + Improvements.

pierre-rouleau:
In the test file:

* Add the `with-mock-forbidding-prompt` macro to replace `with-mock` in tests that execute a dumb-jump command that may prompt when several candidates are found.
  - Replaced the use of `with-mock` by `with-mock-no-prompt` in the appropriate tests.
  - I tested the detection locally in making a code change that would cause prompting and now the test did not hang but simply failed reporting the unexpected forbidden prompting.
  - Add tests to ensure that the `with-mock-forbidding-prompt` macro actually detects prompts and also ensure that it does not interpret another exception for a forbidden prompt.

* Also fixed a spelling mistake typo in the name of a variable named dump-jump instead of dumb-jump (for some reason my finger have hard time typing 'dumb'... now its part of my automatic abbreviations!)

* Add tests for `dumb-jump-use-space-bracket-exp-for`.

* Minor indentation adjustment.

In dumb-jump.el:

* Updated `dumb-jump-find-rules` docstring: it describes the special template conventions, the 2 URls are now formatted as links and updated.

* Renamed the `proj` argument of several functions to `proj-root` to better describe their intent and name them like the others.

* Renamed the `prompt` argument of several functions to `entered-name` to better represent their purpose which is to hold the name of a identifier that was added manually by the user (as a result of a prompt).

* Modified `dumb-jump-process-results` to fix an issue when the line numbers of 2 candidates from different files are located at the same line number. The sorting was not done in that case.  The sorting comparison is now done by a new utility function `dumb-jump--candidate-x<y` instead of duplicate lambda logic (there was no need for closure there).

* Minor indentation and comment.

## Summary by CodeRabbit

* **Refactor**
  * Unified handling of user-entered names across jump/navigation flows for consistent lookups, selection and sorting.
  * Displayed file paths now omit the project root for clearer, shorter paths.
  * Added a stable comparator for consistent result ordering and improved behavior across selector UIs and xref integration.

* **Tests**
  * Added a test macro to forbid unexpected prompts and updated many tests to use it.
  * Renamed an expected-regexps constant and adjusted tests accordingly.

### DIFF
--- a/test/dumb-jump-test.el
+++ b/test/dumb-jump-test.el
@@ -74,20 +74,11 @@
 (defmacro with-mock-forbidding-prompt (&rest body)
   "Test BODY.  Ensure no prompting occurs; fail test if prompting did occur."
   (declare (indent 0))
-  `(let ((have-detected-prompting nil))
-     (condition-case err
-         (with-mock
-           (stub
-            dumb-jump-prompt-user-for-choice
-            => (progn
-                 (setq have-detected-prompting t)
-                 (error "Unexpected call to dumb-jump-prompt-user-for-choice")))
-           ,@body)
-       (error
-        ;; On prompt detection, remember it.
-        (unless have-detected-prompting
-          ;; Another exception was detected; propagate it.
-          (signal (car err) (cdr err)))))))
+  `(with-mock
+     (stub
+      dumb-jump-prompt-user-for-choice
+      => (error "Unexpected call to dumb-jump-prompt-user-for-choice"))
+     ,@body))
 
 ;; ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


pierre-rouleau:
CodeRabbitAi did find a problem I had already fixed in a branch but failed to integrate in my commit.  I will do that next.
There are 2 issues in the macro code: 
- The message string uses the docstring-style `function' quoting.  Unfortunately Emacs processes that during exception generation and changes the format, so the string check fails.
- The if form else part is not inside the if.

I'll fix it in the next commit.

Issue #546, opened 2026-03-09
Code review for NonGNU ELPA


phikal:
Hi, I took some time to look over the code before adding the package to NonGNU ELPA, and I just wanted to post my review with some suggestions here:

```diff
diff --git a/dumb-jump.el b/dumb-jump.el
index d39aa3d274..5a655077db 100644
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -1,9 +1,12 @@
-;;; dumb-jump.el --- Jump to definition for 60+ languages without configuration -*- lexical-binding: t; -*-
+;;; dumb-jump.el --- Jump to definition without configuration -*- lexical-binding: t; -*-
+
 ;; Copyright (C) 2015-2026 jack angers
+
+;; Maintainer: jack angers <jacktasia@gmail.com>
 ;; Author: jack angers and contributors
-;; Url: https://github.com/jacktasia/dumb-jump
+;; URL: https://github.com/jacktasia/dumb-jump
 ;; Version: 0.5.5
-;; Package-Requires: ((emacs "26.1"))
+;; Package-Requires: ((emacs "26.1") (compat "30.1"))
 ;; Keywords: programming
 
 ;; Dumb Jump is free software; you can redistribute it and/or modify it
@@ -21,9 +24,11 @@
 
 ;;; Commentary:
 
-;; Dumb Jump is an Emacs "jump to definition" package with support for 60+ programming languages that favors
-;; "just working" over speed or accuracy.  This means minimal -- and ideally zero -- configuration with absolutely
-;; no stored indexes (TAGS) or persistent background processes.
+;; Dumb Jump is an Emacs "jump to definition" package with support for
+;; 60+ programming languages that favors "just working" over speed or
+;; accuracy.  This means minimal -- and ideally zero -- configuration
+;; with absolutely no stored indexes (TAGS) or persistent background
+;; processes.
 ;;
 ;; Dumb Jump provides a xref-based interface for jumping to
 ;; definitions.  It is based on tools such as grep, the silver searcher
@@ -37,8 +42,8 @@
 ;;
 ;; Now pressing M-.  on an identifier should open a buffer at the place
 ;; where it is defined, or a list of candidates if uncertain.  This
-;; list can be navigated using M-g M-n (next-error) and M-g M-p
-;; (previous-error).
+;; list can be navigated using \\[next-error] and \\[previous-error].
+
 ;;; Code:
 (require 'xref)
 (require 'cl-generic)
[...]

PR #547, opened 2026-03-12
NonGNU ELPA 

jacktasia:
based on #546 

## Summary by CodeRabbit

* **Refactor**
  * Modernized internals using newer Emacs Lisp patterns, reorganized utilities, and improved robustness while preserving user-facing behavior.
  * xref integration and minor-mode declarations updated to use the new patterns.

* **Tests**
  * Updated tests to reflect message-based previews and string-search matching; removed one obsolete test.

* **Chores**
  * Tidied messaging/candidate preview flows and added an ignore manifest for packaging.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -3222,7 +3222,8 @@ Last detected problem is first in the list.")
 
 (defun dumb-jump-env-problem (problem)
   "Add PROBLEM string to `dumb-jump--detected-env-problems' list."
-  (cl-pushnew problem dumb-jump--detected-env-problems :test #'equal))
+  (unless (member problem dumb-jump--detected-env-problems)
+    (push problem dumb-jump--detected-env-problems)))
 
 ;;;###autoload
 (defun dumb-jump-show-detected-problems (keep)
@@ -3931,7 +3932,7 @@ For instance, remove clojure namespace prefix."
     (cl-loop for ent in dumb-jump-language-file-exts
              when (save-excursion
                     (goto-char (point-min))
-                    (search-forward-regexp
+                    (re-search-forward
                      (concat "\\." (regexp-quote (plist-get ent :ext)) "\\b")
                      nil 'noerror))
              collect (plist-get ent :language))))
@@ -4268,9 +4269,8 @@ A language may have more than 1 comment string."
 
 (defun dumb-jump--context-startswith-comment (context-string comments)
   "Return t when CONTEXT-STRING starts with any COMMENTS string."
-  (string-match-p
-   (rx bol (* space) (regexp (regexp-opt comments)))
-   context-string))
+  (let ((cs (string-trim-left context-string)))
+    (cl-some (lambda (c) (string-prefix-p c cs)) comments)))
 
 (defun dumb-jump-filter-no-start-comments (results lang)
   "Filter out RESULTS with a :context starting with a LANG comment in file."
@@ -4757,10 +4757,8 @@ Return a list of (path line context)."
           (save-match-data
             (and (string-match "\\(?:^\\|:\\)\\([0-9]+\\):" resp-line)
                  (match-string 1 resp-line)))))
-    
+
     (cond
-     ;; fixes rare bug where context is blank  but file is defined "/somepath/file.txt:14:"
-     ;; OR: (and (= (length parts) 1) (file-name-exists (nth 0 parts)))
      ((string-match-p ":[0-9]+:$" resp-line)
       nil)
      ((and parts line-num)
@@ -4794,10 +4792,10 @@ Using CUR-FILE and CUR-LINE-NUM to exclude jump origin."
                      (when it
                        (let* ((line-num (string-to-number (nth 1 it)))
                               (diff (- cur-line-num line-num)))
-                         (list (list :path (nth 0 it)
-                                     :line line-num
-                                     :context (nth 2 it)
-                                     :diff diff)))))
+                         (list `(:path ,(nth 0 it)
+                                       :line ,line-num
+                                       :context ,(nth 2 it)
+                                       :diff ,diff)))))
                    parsed)))
     (seq-filter
      (lambda (it)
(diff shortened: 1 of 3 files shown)


dumb-jump.el
```diff
@@ -5607,7 +5475,7 @@ For enabling globally, use `dumb-jump-mode' instead."
            (dumb-jump-message "No symbol under point."))
           ((string-suffix-p " file" lang)
            (dumb-jump-message "Could not find rules for '%s'." lang))
-          ((= (length results) 0)
+          ((length= results 0)
```
phikal: As suggested by @pierre-rouleau, you can use `null` here.

dumb-jump.el
```diff
@@ -5093,14 +4962,14 @@ The arguments are:
      (dumb-jump-populate-regex it look-for variant))
    regexes))
 
-(defun dumb-jump-pcre-to-emacs (regex)
+(defun dumb-jump-pcre-to-emacs (regex)  ;https://github.com/joddie/pcre2el
```
phikal, line 4964: I should have elaborated the comment here, it is currently ambiguous.  The point is that it might be interesting to look into having a hard or soft dependency on pcre2el, unless this function already takes care of everything one needs when translating PCRE to Emacs regular expressions.

dumb-jump.el
```diff
@@ -4877,27 +4744,25 @@ The parameters are:
       (let ((results (funcall parse-fn rawresults cur-file line-num))
             (ignore-case (member lang dumb-jump--case-insensitive-languages)))
         (seq-filter (lambda (it)
-                      (dumb-jump--contains-p look-for
-                                             (plist-get it :context) ignore-case))
+                      (string-search look-for
+                                     (plist-get it :context) ignore-case))
```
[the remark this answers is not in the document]
phikal: Sorry, I missed that.  You could also just always use `string-match-p` with a `regexp-quote`'ed argument, where `case-fold-search` is set to `ignore-case`.

dumb-jump.el
```diff
@@ -3834,12 +3709,12 @@ return a new proplist.  The new proplis is PROPLIS
...
-    (dumb-jump--splice
-     (lambda (it) (string= lang (plist-get it :language)))
-     (lambda (it) (list it (plist-put (copy-tree it) :language newlang)))
-     proplist)))
+  (unless (seq-some (lambda (it) (string= newlang (plist-get it :language)))
+                    proplist)
+    (cl-loop for ent in proplist
+             when (string= (plist-get ent :language) lang)
+             collect (plist-put (copy-tree ent) :language newlang)
+             collect ent)))
```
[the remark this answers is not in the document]
phikal: Please note that we have two `collect` clauses, where one is guarded by a `when`.   This means that the resulting list should already contain all elements in `proplist`.  As a isolated POC, consider

```
(cl-loop for i to 10
         when (cl-oddp i)
         collect (- i)
         collect i)
;=> (0 -1 1 2 -3 3 4 -5 5 6 -7 7 8 -9 9 10)
```

dumb-jump.el
```diff
@@ -4391,21 +4265,12 @@ A language may have more than 1 comment string."
...
-        ;; trim leading white-space from context-string
-        (context-str (replace-regexp-in-string "^\\s-+" "" context-string)))
-    (catch 'found-one
-      (dolist (comment comments)
-        (when (string-prefix-p comment context-str)
-          (setq startswith-comment t)
-          (throw 'found-one nil))))
-    startswith-comment))
+  (let ((cs (string-trim-left context-string)))
+    (cl-some (lambda (c) (string-prefix-p c cs)) comments)))
```
phikal: IIRC my suggestion was to merge comments into a single regular expression and run that against comments, instead of iterating over `comments`.  That way you spend more continuous time in the regular expression engine, and less time invoking closures.

dumb-jump.el
```diff
@@ -5535,23 +5404,22 @@ For enabling globally, use `dumb-jump-mode' instead."
...
-  :global t
-  :keymap dumb-jump-mode-map)
-
+  :global t)
 
 ;; -----------------------------------------------------------------------------
 ;;; Xref Backend
 
-(unless dumb-jump-disable-obsolete-warnings
+(unless (equal dumb-jump-disable-obsolete-warnings (package-get-version))
```
phikal: Can you explicitly confirm that you are OK with this change?  If so, it would make sense to also declare some date by which you will drop support for the old commands.

phikal (commented):
Thank you for taking a look at my proposals, hope these comments help clarify a few things.

dumb-jump.el
```diff
@@ -4891,11 +4887,11 @@ that are both strings."
...
-         (use-ctx (= (seq-count
-                      (lambda (it)
-                        (string= (plist-get it :type)
-                                 (and usable-ctxs (plist-get (car usable-ctxs) :type))))
-                      usable-ctxs)
+         (use-ctx (= (length (seq-filter
+                              (lambda (it)
+                                (string= (plist-get it :type)
+                                         (and usable-ctxs (plist-get (car usable-ctxs) :type))))
+                              usable-ctxs))
```
phikal: Why was this reverted?

dumb-jump.el
```diff
@@ -5076,18 +5071,17 @@ The arguments are:
...
+                                (mapconcat (lambda (it) (format "\\.%s" it)) lang-exts "|")
                                 ")$'"))))
          (exclude-args (dumb-jump-arg-joiner
                         "--ignore-dir"
                         (mapcar
                          (lambda (it)
                            (shell-quote-argument (dumb-jump--replace proj-dir "" it)))
                          exclude-paths)))
-         (regex-args (shell-quote-argument (string-join filled-regexes "|"))))
+         (regex-args (shell-quote-argument (mapconcat #'identity filled-regexes "|"))))
```
phikal: The definition of `string-join` is `mapconcat` with `identity`.

dumb-jump.el
```diff
@@ -5535,23 +5404,22 @@ For enabling globally, use `dumb-jump-mode' instead."
 (define-minor-mode dumb-jump-mode
   "Minor mode for jumping to variable and function definitions."
   ;; Note: independent of the Emacs xref interface.
-  :global t
-  :keymap dumb-jump-mode-map)
-
+  :global t)
```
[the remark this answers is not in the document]
phikal: A minor mode FOO-mode defaults to using FOO-mode-map, if `:keymap` is not specified.

dumb-jump.el
```diff
@@ -5436,7 +5430,7 @@ For enabling globally, use `dumb-jump-mode' instead."
                  "2020-06-26"))
 
 (cl-defmethod xref-backend-identifier-at-point ((_backend (eql dumb-jump)))
-  (and-let* ((bounds (bounds-of-thing-at-point 'symbol)))
+  (when-let* ((bounds (bounds-of-thing-at-point 'symbol)))
```
phikal: It is a common convention to use `and` when the return value is of interest, and `when` when the side effects are more important.  In this case, the former is the case, so it makes sense to hint to this fact.

dumb-jump.el
```diff
@@ -4794,10 +4792,10 @@ Using CUR-FILE and CUR-LINE-NUM to exclude jump origin."
...
                        (let* ((line-num (string-to-number (nth 1 it)))
                               (diff (- cur-line-num line-num)))
-                         (list (list :path (nth 0 it)
-                                     :line line-num
-                                     :context (nth 2 it)
-                                     :diff diff)))))
+                         (list `(:path ,(nth 0 it)
+                                       :line ,line-num
+                                       :context ,(nth 2 it)
+                                       :diff ,diff)))))
```
phikal: Note that in this quasi-quoted list, all non-quoted arguments are self-evaluating, which is why I suggested simplifying it to just `list`.  If you prefer a quasi-quoted list, you can also pull the `list` function call around it into the quasi-quoted code:

```
(list `(...)) = `((...))
```

phikal (commented):
A few more comments on things that confused me.

PR #548, opened 2026-03-16
Fix compiler warnings

pierre-rouleau:
* The `dumb-jump-extra-search-paths-function` docstring has a long line that triggers the docstring-wide warning.  Since it's difficult to refactor the example text, I disabled the compiler docstring-wide around the defcustom explicitly:
  - I started using `(let ((byte-compile-warnings '(not docstrings-wide))) )` block to surround the defcustom form, but that only work in Emacs 30 and later.
  - So I ended up using a with-no-warnings wrapping instead. This will unfortunately disable all warnings inside that code... Maybe you can find a way to express the example in

* The `dumb-jump-get-project-root` either call the new `project-root` if available or `project-roots` in Emacs 26.x.  However that triggers a warning when compiled in later Emacs versions. I surround the code in a `with-no-warnings` form to prevent the warning.

* In `dumb-jump-find-references`: prevent warning about `dump-jump-go` as in the other functions using it.

* Reduced the length of the `dumb-jump-populate-regex-for-emacs` doctring first line.

* Also removed the trailing spaces.

I would recommend to byte-compile the code with warnings treated as errors now that warnings were eliminated from the code base.

## Summary by CodeRabbit

* **Documentation**
  * Clarified and improved docstrings for regex and matching-related functionality.

* **Style**
  * Minor formatting and whitespace refinements.

* **Bug Fixes**
  * Suppressed spurious warning messages on older editor versions to reduce console noise without changing behavior.

### DIFF
--- a/dumb-jump.el
+++ b/dumb-jump.el
@@ -1565,9 +1565,9 @@ If nil add also the language type of current src block."
            :tests ("contract test{"
                    "contract test {"
                    "contract test is foo"))
-    
-    (:language "solidity" :type "type" 
-           :supports ("ag" "grep" "rg" "git-grep") 
+
+    (:language "solidity" :type "type"
+           :supports ("ag" "grep" "rg" "git-grep")
            :regex  "struct\\s*JJJ\\s*\\\{"
            :tests ("struct test{"
                    "struct test {"))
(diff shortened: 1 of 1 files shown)


pierre-rouleau:
@jacktasia  Do you have time to merge this in?

jacktasia (approved):

pierre-rouleau:
@jacktasia   Thank you!

